PR #187, opened 2026-09-03
fix: clear stuck status-bar messages after actions complete

vyrti:
Progress flashes like "deleting …" in bottom status bar never resolved on success and stayed stuck. They now report completion and auto-expire after 8s.

### DIFF
--- a/src/app/actions.rs
+++ b/src/app/actions.rs
@@ -2053,17 +2053,21 @@ impl App {
         self.flash_err = true;
     }
 
-    /// Auto-clear the flash 8s after it last changed. Called from the main
-    /// loop's 1s tick; detects a change by diffing against `flash_seen`
-    /// rather than requiring every `self.flash = …` call site to also touch
-    /// a timestamp.
+    /// Auto-clear a successful transient flash 8s after it last changed.
+    /// Errors and the initial welcome hint remain until another interaction
+    /// replaces them. Called from the main loop's 1s tick; detects a change by
+    /// diffing against `flash_seen` rather than requiring every
+    /// `self.flash = …` call site to also touch a timestamp.
     pub fn expire_flash(&mut self) {
         if self.flash != self.flash_seen {
             self.flash_seen = self.flash.clone();
             self.flash_since = std::time::Instant::now();
             return;
         }
-        if !self.flash.is_empty() && self.flash_since.elapsed() >= std::time::Duration::from_secs(8)
+        if !self.flash.is_empty()
+            && !self.flash_err
+            && self.flash != WELCOME_FLASH
+            && self.flash_since.elapsed() >= std::time::Duration::from_secs(8)
         {
             self.flash.clear();
             self.flash_err = false;
--- a/src/app/actions.rs
+++ b/src/app/actions.rs
@@ -91,7 +91,7 @@ impl App {
         let tx = self.tx.clone();
         let genr = self.generation;
         tokio::spawn(async move {
-            let mut errors = 0u32;
+            let mut failed = false;
             for (name, ns) in targets {
                 let api: Api<DynamicObject> = if kind.namespaced && !ns.is_empty() {
                     Api::namespaced_with(client.clone(), &ns, &kind.ar)
@@ -99,7 +99,7 @@ impl App {
                     Api::all_with(client.clone(), &kind.ar)
                 };
                 if let Err(e) = api.patch(&name, &PatchParams::default(), &patch).await {
-                    errors += 1;
+                    failed = true;
                     let _ = tx
                         .send(Msg::Error {
                             generation: genr,
@@ -108,12 +108,12 @@ impl App {
                         .await;
                 }
             }
-            if errors == 0 {
+            if !failed {
                 let _ = tx
                     .send(Msg::Flash {
                         generation: genr,
                         message: ok_message,
-                        ok: true,
+                        err: false,
                     })
                     .await;
             }
@@ -153,7 +153,7 @@ impl App {
             if force {
                 dp = dp.grace_period(0);
             }
-            let mut errors = 0u32;
+            let mut failed = false;
             for (name, ns) in targets {
                 let api: Api<DynamicObject> = if kind.namespaced && !ns.is_empty() {
                     Api::namespaced_with(client.clone(), &ns, &kind.ar)
@@ -161,7 +161,7 @@ impl App {
                     Api::all_with(client.clone(), &kind.ar)
                 };
                 if let Err(e) = api.delete(&name, &dp).await {
-                    errors += 1;
+                    failed = true;
                     let _ = tx
                         .send(Msg::Error {
                             generation: genr,
@@ -170,12 +170,12 @@ impl App {
                         .await;
                 }
             }
-            if errors == 0 {
+            if !failed {
                 let _ = tx
                     .send(Msg::Flash {
                         generation: genr,
                         message: done_label,
-                        ok: true,
+                        err: false,
                     })
                     .await;
             }
@@ -288,22 +288,25 @@ impl App {
             format!("draining {} nodes…", targets.len())
         };
         self.flash_err = false;
+        // Not "drained": the task is done once the API accepts the evictions,
+        // but the pods are still terminating at that point (real `kubectl
+        // drain` blocks until they're gone).
         let done_label = if targets.len() == 1 {
-            format!("drained {}", targets[0])
+            format!("drain requested: {}", targets[0])
         } else {
-            format!("drained {} nodes", targets.len())
+            format!("drain requested: {} nodes", targets.len())
         };
         tokio::spawn(async move {
             let nodes: Api<DynamicObject> = Api::all_with(client.clone(), &kind.ar);
             let node_patch = Patch::Merge(node_unschedulable_patch(true));
             let pods: Api<Pod> = Api::all(client.clone());
-            let mut errors = 0u32;
+            let mut failed = false;
             for node in targets {
                 if let Err(e) = nodes
                     .patch(&node, &PatchParams::default(), &node_patch)
                     .await
                 {
-                    errors += 1;
+                    failed = true;
                     let _ = tx
                         .send(Msg::Error {
                             generation: genr,
@@ -319,7 +322,7 @@ impl App {
                 let pod_list = match listed {
                     Ok(list) => list,
                     Err(e) => {
-                        errors += 1;
+                        failed = true;
                         let _ = tx
                             .send(Msg::Error {
                                 generation: genr,
@@ -346,7 +349,7 @@ impl App {
                             if let Err(delete_err) =
                                 pod_api.delete(name, &DeleteParams::default()).await
                             {
-                                errors += 1;
+                                failed = true;
                                 let _ = tx
                                     .send(Msg::Error {
                                         generation: genr,
@@ -358,7 +361,7 @@ impl App {
                             }
                         }
                         Err(e) => {
-                            errors += 1;
+                            failed = true;
                             let _ = tx
                                 .send(Msg::Error {
                                     generation: genr,
@@ -369,12 +372,12 @@ impl App {
                     }
                 }
             }
-            if errors == 0 {
+            if !failed {
                 let _ = tx
                     .send(Msg::Flash {
                         generation: genr,
                         message: done_label,
-                        ok: true,
+                        err: false,
                     })
                     .await;
             }
@@ -763,7 +766,7 @@ impl App {
             format!("set-image {container}={image}"),
             format!("{name} in {ns}"),
         );
-        self.flash = format!("setting image: {container} → {image}");
+        self.flash = format!("setting image: {container} → {image}…");
         self.flash_err = false;
         let ok_message = format!("image set: {container} → {image}");
         self.spawn_patch_action(
@@ -1539,10 +1542,10 @@ impl App {
         let label = self.action_label(&targets);
         self.note_action(format!("scale to {replicas}"), label);
         self.flash = if let [(name, _)] = targets.as_slice() {
-            format!("scaling {name} → {replicas}")
+            format!("scaling {name} → {replicas}…")
         } else {
             format!(
-                "scaling {} {} → {replicas}",
+                "scaling {} {} → {replicas}…",
                 targets.len(),
                 self.kind_plural
             )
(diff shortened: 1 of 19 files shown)


nklmilojevic (changes requested):
Nice catch on this one, the `Msg::Flash` + tick approach is the right shape and the `flash_seen` diffing is clever - auto-times all ~90 `self.flash = ...` call sites without touching any of them. Partial failures look correct too, I checked each `errors == 0` guard and a success flash can't overwrite an error.

Two things before merge though.

**The welcome banner now disappears after 8s.** `App::new` sets `flash` to the "Welcome to sofka - ':' resource, enter drill, d describe, l logs, ? help" hint, and `expire_flash` doesn't know it's special, so it clears on the eighth tick. Verified rather than guessed:

```rust
let (mut app, _rx) = test_app();
assert!(app.flash.starts_with("Welcome to sofka"));
app.expire_flash();
app.flash_since = std::time::Instant::now() - std::time::Duration::from_secs(9);
app.expire_flash();
// app.flash == ""
```

That's the one string a first-time user most needs to stay put, so it shouldn't expire.

**Errors expire too.** Same setup with `flash_err = true` and the message is gone after 9s, flag reset. Scope here is stuck *progress* messages, and a failed delete self-erasing is a different thing - if you step away for a minute you come back to a clean bar and no idea anything broke. I'd gate the expiry on `!self.flash_err` so errors stick until the next action replaces them.

Both are easy to cover with tests, and there's already `detail_arrival_clears_progress_flash` right next door to model them on. Would be good to have them here since this adds a new `Msg` variant and a timing rule.

Minor, not blocking: `switching to {name}...` in `pickers.rs:678` is still an unresolved progress flash, so the bug class isn't fully closed. Fine as a follow-up.

Tnx!

src/app/actions.rs
```diff
@@ -1948,6 +2053,28 @@ impl App {
...
+    pub fn expire_flash(&mut self) {
+        if self.flash != self.flash_seen {
+            self.flash_seen = self.flash.clone();
+            self.flash_since = std::time::Instant::now();
+            return;
+        }
+        if !self.flash.is_empty()
+            && !self.flash_err
+            && self.flash != WELCOME_FLASH
+            && self.flash_since.elapsed() >= std::time::Duration::from_secs(8)
```
nklmilojevic: This wipes in-flight progress flashes too, not just finished ones. Any non-error, non-welcome flash goes after 8s, including the `"draining 3 nodes…"` message this is supposed to complement. A drain or a big multi-target delete easily runs longer than that, so the UI goes progress -> blank -> `"drained 3 nodes"`. The blank stretch looks like nothing is happening, which is worse than the bug you're fixing.

We already end every progress flash with `…`, so adding `&& !self.flash.ends_with('…')` here is a one-liner that fits the convention.

src/store.rs
```diff
@@ -214,6 +214,13 @@ pub enum Msg {
         generation: u64,
         error: String,
     },
+    /// A background action (delete, restart, scale, drain, helm op, …)
+    /// finished; replaces its "…ing" progress flash with a result.
+    Flash {
+        generation: u64,
+        message: String,
+        ok: bool,
```
nklmilojevic: `ok` is dead. All six construction sites pass `ok: true` (`actions.rs:116,178,377,1985,2161,2242`) and failures go through `Msg::Error`, so nothing can ever send false. Drop the field and set `flash_err = false` in the handler.

src/app/actions.rs
```diff
@@ -248,15 +288,22 @@ impl App {
             format!("draining {} nodes…", targets.len())
         };
         self.flash_err = false;
+        let done_label = if targets.len() == 1 {
+            format!("drained {}", targets[0])
+        } else {
+            format!("drained {} nodes", targets.len())
+        };
```
nklmilojevic, line 306: `"drained {node}"` isn't true yet at this point. The task reports success once evictions are accepted, pods are still terminating. Real `kubectl drain` blocks until they're gone. Something like `"drain requested: node-1"` is honest.

`"deleted web"` on the delete path is fine by comparison, the watch drops the row right after.

src/app/lifecycle.rs
```diff
@@ -677,6 +677,14 @@ impl App {
...
                 self.flash_err = true;
             }
+            Msg::Flash {
+                generation,
+                message,
+                ok,
+            } if generation == self.generation => {
+                self.flash = message;
+                self.flash_err = !ok;
+            }
```
nklmilojevic, line 849: `Msg::CanIResult { generation, text, ok }` at `lifecycle.rs:822` is now the exact same shape with an identical handler. Would be nice to fold it in and have one message instead of two.

src/app/actions.rs
```diff
@@ -90,13 +91,15 @@ impl App {
         let tx = self.tx.clone();
         let genr = self.generation;
         tokio::spawn(async move {
+            let mut errors = 0u32;
```
nklmilojevic: Only ever compared to zero, here and in the other three loops. A bool reads better.

src/app/mod.rs
```diff
@@ -1186,6 +1191,11 @@ pub struct App {
     pub cmd_sel: usize,
     pub flash: String,
     pub flash_err: bool,
+    /// Last flash text observed by [`App::expire_flash`], so a change can be
+    /// detected (and re-timestamped) without touching every call site that
+    /// sets `flash` directly.
+    pub(super) flash_seen: String,
+    pub(super) flash_since: std::time::Instant,
```
nklmilojevic, line 1230: Two small notes on this mechanism, neither blocking.

The `flash_seen` diffing means repeating the same action inside the window doesn't reset the timer, so `:ns foo` twice can make the second confirmation vanish a second later. Edge case, but it's invisible from the ~90 call sites so someone will hit it eventually.

And matching `WELCOME_FLASH` by string works, but a `flash_sticky: bool` would survive someone editing that constant. Low priority, the diffing approach is a fine trade against touching every call site.

nklmilojevic (changes requested):
Three things I'd like changed before merge, left inline: the 8s expiry wiping in-flight progress flashes, the dead `ok` field, and the drain message claiming more than it did. The rest are smaller notes, take them or leave them.

Everything else checked out. Generation gating follows the `Msg::Error` precedent and every bump (`lifecycle.rs:251`, `bump_generation`) is a view switch that sets its own flash. `run_helm` returns `Err` on non-zero exit so the rollback flash can't fire on a failed run. Losing the informational flashes to the timer is fine, the header and table title already carry ns, context and filter :)

vyrti:
Pushed 8d722f3. Both Greptile P1s were real, so they are fixed with tests:

**Result handler erasing unrelated status.** Reachable: `explain.rs:28` and `gitops.rs:33` set `mode` before spawning, and `Msg::Detail` leaves you on the table while the describe is in flight. Start a delete, open describe, delete lands first — `self.flash.clear()` wiped the confirmation. With an error instead it is worse, since errors no longer expire, so that was permanent loss. Clearing is now conditional on the bar still showing an in-flight `…` message. That turned out to be the same rule the generation-bump path already used, so the two helpers collapsed into one `clear_progress_flash()`.

**Older action overwriting newer status.** Mostly last-writer-wins on a single status line, and the messages are all true. The one real defect is a late success erasing a *displayed error*. `Msg::Flash` now yields to an error on the bar; starting the next action clears `flash_err` through its own progress flash, so the ordinary one-action flow is untouched. I did not add per-action identity to `Msg::Flash` for the progress-overwrite half — you explicitly blessed the generation-gating shape, so that felt like your call rather than mine.

Sweeping for the same classes turned up five more, all in this branch:

- `"setting image: c → i"` and `"scaling web → 3"` had no trailing `…`, so the new guard blanked them mid-patch. The convention was not actually universal. `action_progress_flashes_keep_the_ellipsis_convention` now drives all six patch actions and asserts it.
- `"explaining X…"` was never cleared — `Msg::Explain` had no flash handling at all, unlike `Msg::Gitops` ten lines below it. Stuck forever under the new guard.
- `"reconciled X"` and `"refreshed X"` are the same overclaim as `drained`: both only stamp an annotation (`reconcile.fluxcd.io/requestedAt`, `force-sync`) and the controller acts later. Now `"reconcile requested:"` / `"refresh requested:"`.
- `key_flux_menu` recomputes `action_targets()` at Enter, but only `request_flux_menu` checks it. If the row is deleted while the menu is open you got `"suspended 0 kustomizations"` from a zero-iteration loop. Guarded.

Two of your notes I resolved in a way you may want to reverse:

1. `ok` vs folding in `CanIResult` conflict — folding makes the field live, because a `:can-i` denial needs error styling but is not a `Msg::Error` (that would bump `watch_errors` and prefix `"error: "`). Kept as `err: bool` on the merged variant. Happy to split them back if you prefer the field gone.
2. Gating expiry on `!flash_err` pins benign hints too. `flash_warn` serves both real failures and throwaway feedback like `"invalid replica count"` or `"no owner reference"`, which now sit there until the next action. Followed your instruction literally rather than reclassifying ~40 call sites unprompted, but splitting the two is probably the right follow-up.

530 tests pass, fmt and clippy -D warnings clean. I reverted each P1 fix locally to confirm the new tests fail without them.

src/app/actions.rs
```diff
@@ -1948,6 +2075,69 @@ impl App {
...
+    /// refresh (`r`) during a bulk delete strands that message for good. And a
+    /// document/report task finishing (`Msg::Detail`/`Explain`/`Gitops`) is
+    /// done with its own "describing web…", but a concurrent action may have
+    /// claimed the bar since — clearing unconditionally would wipe that
+    /// action's confirmation, or a sticky error, and leave nothing behind.
+    pub(super) fn clear_progress_flash(&mut self) {
+        if self.flash.ends_with('…') {
+            self.flash.clear();
+            self.flash_err = false;
+        }
```
nklmilojevic, line 2251: This still clears a concurrent action's progress because the ellipsis identifies only the message type, not the operation that owns it. For example: describe sets `describing web…`, delete then replaces it with `deleting web…`, and when the describe result arrives its `Msg::Detail` handler calls this helper and clears the delete progress while that delete is still running. The new concurrent-result test does not cover this ordering. Please give status claims an operation token and clear/replace only when the completing task still owns the bar. The same ownership mechanism should address Greptile's related finding where an older action result overwrites a newer action's progress or result.

nklmilojevic (changes requested):

nklmilojevic:
just one more change and we are ready to go

vyrti:
Pushed 44828f3, fixing two gaps I found auditing the `StatusClaim` work.

**A failure could vanish entirely.** Moving action failures from `Msg::Error` onto claimed `Msg::Flash` made them subject to ownership, so a delete that failed while a newer operation held the bar was dropped outright. Verified rather than reasoned about:

```
delete claims "deleting web…", describe then claims "describing api…", delete fails →
flash = "describing api…"   flash_err = false   last_error = None   watch_errors = 0
```

Nothing on the bar, nothing in `:info`. That is the case you raised in the first review — step away, come back to a clean bar, no idea anything broke.

The rule now: a failure is always recorded in `last_action_error` and reported by `:info`, so it can never be lost outright. It may also borrow the bar while the owner has only an unresolved `…` showing, since an in-progress message is not news anyone misses. It never displaces the owner's finished result — `an_older_action_result_cannot_overwrite_a_newer_action` still passes unchanged — and borrowing does not transfer ownership, so the owner still reports normally when it lands.

**Twelve operations never got claims.** Same ordering bug the token exists to fix. With a delete's `deleting 3 pods…` owning the bar:

```
older Msg::FindResults   → "0 hit(s) for 'foo'"
older Msg::TransferDone  → "copied a -> b"
older Msg::SnapshotSaved → "saved snapshot → /tmp/snap.json"
```

`:find`, `:bundle`, `:bundle-save`, `:snapshot`, kubectl cp, node-debugger cleanup, plugins (popup and background), pulse, xray, log save, clipboard copy and context rename all claim now — 31 claim sites, up from 18. Several had no progress message to own, so they gained one (`saving bundle…`, `saving logs…`, `copying to clipboard…`, `renaming X → Y…`).

Left unclaimed deliberately: synchronous view switches and filters, which take the bar directly and invalidate the claim — that is the documented behaviour and `an_unclaimed_status_update_invalidates_the_current_claim` pins it. Also `Msg::Error`, `Msg::Panic` and `Msg::Notify`, which are watch- or process-level and must surface from any view, and `Msg::ContextSwitched`, which bumps the generation and rebuilds the world anyway.

Two new tests, `an_action_failure_is_never_silently_dropped` and `every_async_result_is_scoped_to_its_own_operation`. I reverted each fix locally to confirm both actually fail without them. 533 pass, fmt and clippy -D warnings clean.

nklmilojevic (changes requested):
Requesting changes for one thing.

`set_claimed_status` drops an error result whenever a newer operation has claimed the bar. The second half of `an_older_action_result_cannot_overwrite_a_newer_action` codifies it: "delete api failed: forbidden" is never shown. On main that went through `Msg::Error`, so it was always displayed and recorded in `last_error`; now it reaches neither and the failure leaves no trace.

Let `err: true` bypass the ownership check (a failure outranks a newer confirmation) and set `last_error` in the `Msg::Flash` arm. Leave `watch_errors` alone, keeping action failures out of that counter is an improvement.

nklmilojevic:
Thanks for the thorough follow-ups here - the claim-based direction looks good, but I found a few remaining blockers, so I do not think this is ready to merge yet:

- Pulse/Xray reuse one claim across recurring polls. After a successful poll clears it, later incomplete results can no longer show their warning, leaving incomplete dashboard data unmarked.
- Saving logs uses `log_gen`. If the user exits Logs before the write completes, the result is dropped while `saving logs…` remains stuck indefinitely.
- Background notifications/errors can still invalidate an action claim, causing its eventual completion to be discarded.

The existing suite passes (544 passed, 1 ignored), but targeted regression tests reproduce the Pulse and log-save cases. Could you address these paths and add coverage for them?

nklmilojevic (changes requested):
Requesting changes based on the blocking issues described in my previous comment.

nklmilojevic:
I pushed `24f05d5` to close the final status-ownership race.

When an older action failed while a newer report was still loading, the failure borrowed the report's progress claim. The report's successful completion then cleared the shared status bar, which erased the sticky action failure. `clear_claimed_status` now relinquishes the completed report's claim without clearing a displayed error.

The regression test now verifies that the report still opens, the action failure remains visible, and the completed claim is released.

Validation: 547 tests passed (1 ignored), formatting is clean, and Clippy passes with warnings denied.

PR #188, opened 2026-09-03
perf: cheaper filters, Helm decode and caches; watch node pod counts

vyrti:
Follows #185 with the next tier of hot-path work. Measured with
`benches/hot_paths.rs` on 2,000 synthetic pods / 300 Helm releases / 128
contexts, Apple Silicon, bench profile. Every figure in the table below is
against `main`. Four bench groups are new (`filter_cmp`, `helm_rows`,
`fleet_marks`, `helm_decode`) — the structured filter and Helm dedup paths
had no coverage at all.

## Results

| Benchmark | Before | After | |
| --- | ---: | ---: | ---: |
| context switcher fleet marks (128 contexts) | 202 µs | **26 µs** | −87% |
| Helm list rebuild, filter keystroke | 35.7 µs | **11.0 µs** | −70% |
| filter `namespace=…` | 157 µs | **64 µs** | −60% |
| filter `status=…` | 767 µs | **671 µs** | −17% |
| filter, no match | 225 µs | **207 µs** | −14% |
| cells, helm (16 rows) | 387 µs | **336 µs** | −13% |
| filter, broad | 489 µs | **467 µs** | −7% |
| filter, name hit | 609 µs | **582 µs** | −7% |
| row update + redraw query | 9.3 µs | **8.8 µs** | −5% |
| cells, pods (256 rows) | 298 µs | **282 µs** | −4% |

## What changed

- **Nodes view watches pods instead of re-listing every 10s.** Counts are kept
  incrementally and coalesced to at most one message per second. Publishing is
  gated on `InitDone` — the initial list arrives as a stream, so emitting
  mid-init would walk the PODS column up from zero on every resync.
- **Backoff and RBAC fallback on that watch.** The first version retried a
  refused watch as fast as the API server could answer (~9,800 requests/s in
  the mock test). Other transient failures now use client-go's exponential
  backoff, while a forbidden list/watch switches to the old 10-second list
  poll. This keeps counts fresh for list-only RBAC without retrying a denied
  watch.
- **Structured filter comparisons allocate nothing per object.** `eval_cmp`
  lowercased both the cell and the needle for every object on every rebuild;
  the needle is now folded once at parse time and the cell folded lazily
  through an iterator. `column_cell` returns `Cow<str>` so a `namespace=…`
  comparison borrows instead of cloning.
- **The Helm release list decodes only what it renders.** `RawRelease`
  materialized `manifest: String` and `config: Value` — the bulk of a release
  payload, and neither is read by any column. `decode_summary` walks past
  them; the detail views still use the full `decode`.
- **Helm latest-revision dedup cached against a store mutation counter.** It
  rescanned the whole store on every rebuild, including rebuilds staled by a
  filter keystroke or a sort toggle, which cannot change the dedup.
- **`is_fleet_context` answers membership directly** instead of building and
  cloning a `Vec` of every fleet context per row, per frame.
- **Row cache retention is bounded against the store, and hands capacity
  back.** The bound has to be store size, not visible row count: the filter
  warms a cell entry for every object it tests, including the ones it rejects,
  so a narrow filter legitimately leaves far more cells than keys. `retain`
  alone also keeps the peak allocation, and since `invalidate_row` already
  drops deleted rows one at a time, length never signalled the oversized table
  a large view leaves behind — the shrink is gated on capacity, with 4×
  hysteresis so it settles instead of rehashing every rebuild.
- **Timeline history bounded** at 50 entries per object rather than 200.
- `sort_unstable_by` where the comparator is a total order over the element.
  The two sites where it is not — `FindItem::ns` and `Suggestion::kind` sit
  outside their tie-breaks, so an unstable sort would reorder equal-comparing
  items between runs — stay stable and say why.

## Measured and not kept

**SIMD JSON.** The typed parse is 12.1 µs of a 24.1 µs Helm decode, so the
parse genuinely dominates. But the reason is that the parse was materializing
payloads no column reads; skipping those got −13% with no new dependency and
no `unsafe`. A SIMD parser on top would optimize work that no longer happens.

**`panic = "abort"`.** Safe here — no `catch_unwind` anywhere, and the panic
hook that restores the terminal still runs — but cargo reports `` `panic`
setting is ignored for `bench` profile ``, so no speedup could be demonstrated
with this harness. Skipped rather than taken on faith. `codegen-units = 1`
remains rejected from #185.

## Verification

- `cargo fmt --check`
- `cargo clippy --all-targets --all-features` (clean)
- `cargo test --all-features` — 524 passed, 1 existing ignored
- New tests: every `Store` mutation advancing the version the dedup cache is
  keyed on (mutation-checked — removing any one of the seven bumps fails it),
  the dedup refreshing when a new revision arrives, the row cache both
  releasing capacity and then settling, `decode_summary` agreeing with the
  full decode, and a mock API server that changes its list after refusing `watch`,
  asserting the 10-second fallback publishes refreshed counts and stops
  retrying the denied watch.

The version counter is a plain `u64` by design: `RowKey` is `Rc<str>`, so
`Store` is not `Send` and the compiler rejects it crossing a thread boundary.
The counters that do cross one — `gen_flag`, `log_flag`, `streaming_lists`,
the theme epoch — are atomics.

Note: `theme::tests::helm_release_statuses_are_colored` is flaky on `main`
independently of this branch (~1 in 6 runs) — `theme.rs:358` keeps the palette
in a process-global that `ui.rs:3614` mutates from another test. Worth a
separate issue.

### DIFF
--- a/src/app/lifecycle.rs
+++ b/src/app/lifecycle.rs
@@ -569,6 +569,11 @@ impl App {
     /// node resources — mirroring `kubectl describe node`. Replaces a full
     /// cluster-wide pod re-list every 10s with one watch, kept incrementally
     /// up to date and coalesced to at most one `Msg::NodePods` per second.
+    ///
+    /// RBAC granting `list` but not `watch` still works: `watcher` serves the
+    /// initial list itself, so counts publish from that, and a refused watch
+    /// just sends it round again for a fresh list. Backoff is what keeps that
+    /// loop from becoming a hot retry against the API server.
     pub(super) fn spawn_node_pods_poll(&mut self) {
         let Some(pkind) = self.cluster.resolve("pods") else {
             return;
@@ -584,7 +589,12 @@ impl App {
             let cfg = watcher::Config::default()
                 .any_semantic()
                 .fields("status.phase!=Succeeded,status.phase!=Failed");
-            let mut stream = watcher(api, cfg).boxed();
+            // `watcher` re-lists as fast as the stream is polled, so an error
+            // that does not clear itself (a refused watch) would hammer the API
+            // server — measured at ~9,800 requests a second against a test
+            // server. `DefaultBackoff` is client-go's: 800ms doubling to 30s,
+            // jittered, reset after 2 minutes of quiet.
+            let mut stream = watcher(api, cfg).default_backoff().boxed();
             // Node per pod, kept incrementally so per-node counts never need
             // a full rescan of the cluster's pods.
             let mut pod_nodes: HashMap<String, String> = HashMap::new();
@@ -649,6 +659,8 @@ impl App {
                                 synced = true;
                                 dirty = true;
                             }
+                            // The stream self-heals, and the backoff above
+                            // paces the retries.
                             Err(_) => {}
                         }
                     }
--- a/src/app/lifecycle.rs
+++ b/src/app/lifecycle.rs
@@ -590,6 +590,11 @@ impl App {
             let mut pod_nodes: HashMap<String, String> = HashMap::new();
             let mut counts: HashMap<String, usize> = HashMap::new();
             let mut dirty = false;
+            // The initial list arrives as a stream of `InitApply`s, so the
+            // counts are incomplete until `InitDone`. Publishing mid-init
+            // would walk the PODS column up from zero on every (re)sync —
+            // the old full-list poll only ever emitted complete counts.
+            let mut synced = false;
             let mut ticker = tokio::time::interval(Duration::from_secs(1));
             ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
             loop {
@@ -638,9 +643,12 @@ impl App {
                             Ok(watcher::Event::Init) => {
                                 pod_nodes.clear();
                                 counts.clear();
+                                synced = false;
+                            }
+                            Ok(watcher::Event::InitDone) => {
+                                synced = true;
                                 dirty = true;
                             }
-                            Ok(watcher::Event::InitDone) => dirty = true,
                             Err(_) => {}
                         }
                     }
@@ -648,7 +656,7 @@ impl App {
                         if flag.load(Ordering::SeqCst) != genr {
                             break;
                         }
-                        if dirty {
+                        if dirty && synced {
                             if tx
                                 .send(Msg::NodePods {
                                     generation: genr,
--- a/src/app/lifecycle.rs
+++ b/src/app/lifecycle.rs
@@ -564,10 +564,11 @@ impl App {
         self.tasks.push(handle);
     }
 
-    /// Poll the pods API for the nodes view: pod count per node (the PODS
+    /// Watch the pods API for the nodes view: pod count per node (the PODS
     /// column). Counts non-terminated pods — Succeeded/Failed pods hold no
-    /// node resources — mirroring `kubectl describe node`. List failures
-    /// (e.g. no cluster-wide pod list permission) leave the column at "-".
+    /// node resources — mirroring `kubectl describe node`. Replaces a full
+    /// cluster-wide pod re-list every 10s with one watch, kept incrementally
+    /// up to date and coalesced to at most one `Msg::NodePods` per second.
     pub(super) fn spawn_node_pods_poll(&mut self) {
         let Some(pkind) = self.cluster.resolve("pods") else {
             return;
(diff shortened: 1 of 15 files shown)


vyrti:
this will be heavily updated, does not worth a look now

vyrti:
@nklmilojevic extra carefull review needed here, as this PR touch how code gets the the data from cluster and its fallback if this new way does not work correctly (forbidden), but saves tons of resources and traffic if works (both on API and locally)

src/app/lifecycle.rs
```diff
@@ -579,13 +594,128 @@ impl App {
...
+                            // that case. This also paces a refused initial list,
+                            // whose preceding `Init` event would otherwise keep
+                            // resetting `DefaultBackoff` to its minimum delay.
+                            Err(error) if node_pods_watch_forbidden(&error) => {
+                                poll_fallback = true;
+                                break;
+                            }
+                            // The stream self-heals, and the backoff above
+                            // paces other transient failures.
+                            Err(_) => {}
```
nklmilojevic: This does not get exponential backoff for persistent initial-list failures. `kube-runtime` emits `Ok(Event::Init)` before each list attempt, and `default_backoff()` resets on every successful stream item. The sequence is therefore `Init -> 500 -> minimum delay -> Init -> 500 -> minimum delay`, so repeated 500s or transport failures never advance beyond the first interval. I reproduced the same stream sequence locally: retries stayed about 1.2 seconds apart instead of increasing. This regresses the previous 10-second failure pacing and repeatedly sends cluster-wide pod lists while the API server is degraded. Please manage the backoff so it resets only after `InitDone` or an established watch event.

src/app/tests.rs
```diff
@@ -24,6 +24,196 @@ fn alt(code: KeyCode) -> KeyEvent {
...
+    cluster.cluster_url = url.into();
+    cluster
+}
+
+/// RBAC granting `list` but not `watch`. The first list seeds counts, then the
+/// refused watch must switch to periodic lists: kube's watcher retries only
+/// the watch from the same resourceVersion, so it cannot provide that fallback
+/// itself. The mock changes its list response to prove counts remain fresh.
+#[tokio::test]
+async fn node_pods_survives_a_forbidden_watch_without_hammering_the_api() {
```
nklmilojevic, line 283: The forbidden fallback is covered, but the normal incremental path and the non-403 failure path are not. Please add coverage for initial sync, apply, pod reassignment to another node, delete, and `Init`/`InitDone` resync. Also add a repeated initial-list 500 test that verifies the retry intervals increase - that would catch the backoff reset above.

src/helm.rs
```diff
@@ -87,25 +101,76 @@ struct RawMetadata {
...
+}
+
 /// Decode a release Secret into its `Release`. `None` if the secret doesn't
 /// carry a `data.release` payload or it can't be decoded (corrupt, unknown
 /// format, or not a Helm release secret at all) — callers treat that as
 /// "unrenderable", not a crash.
-pub fn decode(secret: &DynamicObject) -> Option<Release> {
+/// The release payload as JSON: `data.release` is base64 twice over, then
+/// gzipped. Shared by [`decode`] and [`decode_summary`].
+fn release_json(secret: &DynamicObject) -> Option<Vec<u8>> {
```
nklmilojevic: The old `decode` documentation now attaches to `release_json()` and says this function returns a `Release`. Please keep only the compressed-payload comment here and move the `Release`/`None` contract back above `decode()`.

nklmilojevic (changes requested):

vyrti:
Pushed 6be913d. All four addressed.

**Backoff (`lifecycle.rs:685`).** You were right, and the mechanism is exactly as you described — confirmed in the 4.2.0 source: `watcher.rs:523` yields `Ok(Event::Init)` from `State::Empty` before the list is attempted, and `stream_backoff.rs:88` resets on any `Ok`. So a failing initial list cycles `Init -> InitialListFailed -> minimum delay` forever.

Pacing is now driven in the loop instead of through `.default_backoff()`. Same strategy — `DefaultBackoff`, so still client-go's 800ms/×2/30s/jittered — but reset only on `Apply`, `Delete` or `InitDone`, never on the replayed `Init`/`InitApply`. The new test reproduces your measurement before the fix: `[1.15292375s, 961.497125ms]`, non-increasing, right about the 1.2s you saw. After it, the gaps grow.

**Tests (`tests.rs:121`).** Added a mock that serves a real list *and* a chunked watch stream the test pushes frames into, so the incremental path is driven event by event rather than inferred:

- `node_pods_counts_follow_the_watch_incrementally` — initial sync (nothing published before `InitDone`), add, reassignment to another node (decrements the old node as well as incrementing the new), delete taking the node out of the map rather than leaving a zero.
- `node_pods_rebuilds_counts_after_a_desync_resync` — a 410 `ERROR` frame desyncs the watch, `watcher` re-lists, and the counts are rebuilt from that list alone. The second list returns a completely different pod set, so a merge instead of a rebuild would strand `node-a`.
- `node_pods_backs_off_when_the_initial_list_keeps_failing` — the non-403 path, asserting the interval grows.

**Unicode folding (`rows.rs:203`, Greek final sigma).** Real, and I checked it rather than assuming — `"ΟΔΟΣ".to_lowercase()` is `"οδος"` while the per-`char` fold gives `"οδοσ"`. Fixed by making both sides share one `fold_lower`, so the needle can't be folded one way at parse time and the cell another way on rebuild. Kept the per-`char` form: `str::to_lowercase` on the cell would allocate per object and undo the point of the change. Regression test in `filter.rs` pins the two sides together. In practice only a CRD printer column or a user-defined view column can carry text like that — namespaces and pod names are DNS labels — so the exposure is narrow, but the fix is free.

**Doc placement (`helm.rs:138`).** Moved. `release_json` keeps only the compressed-payload note; the `Release`/`None` contract is back above `decode()`.

528 tests pass, fmt and clippy -D warnings clean, benches still build.

One unrelated thing worth knowing: `theme::helm_release_statuses_are_colored` failed once for me in a full run and I could not reproduce it in 17 runs after (9 with these changes, 8 on the branch without them). It looks like the global `ACTIVE` palette `RwLock` racing a test that swaps skins, which nothing here touches — flagging it rather than chasing it in this PR.

nklmilojevic:
@vyrti thanks again for this, I really appreciate the effort! One blocker in the Unicode fix to address and we will be able to ship this:

fold_lower() now applies char::to_lowercase to both sides. That fixes identical uppercase strings, but it does not preserve the previous case-insensitive behavior when the casing differs.

I reproduced these results:

- Old behavior: "ΟΔΟΣ".to_lowercase() == "οδος".to_lowercase() is true.
- New behavior: fold_lower("ΟΔΟΣ") == fold_lower("οδος") is false.
- The reverse direction also fails.

The new test compares uppercase "ΟΔΟΣ" with the same uppercase value, so it misses the actual case-insensitive case.

The safest fix is to retain an allocation-free ASCII path and use whole-string str::to_lowercase() whenever either side is non-ASCII. Add mixed-case tests in both directions.

Everything else is clean!

vyrti:
> I reproduced these results:
> 
>     * Old behavior: "ΟΔΟΣ".to_lowercase() == "οδος".to_lowercase() is true.
> 
>     * New behavior: fold_lower("ΟΔΟΣ") == fold_lower("οδος") is false.
> 
>     * The reverse direction also fails.

yeah my second agent spotted that too, thats why PR still draft, on it :)

nklmilojevic (approved):
LGTM :)

PR #189, opened 2026-09-03
feat: let a view configure where `enter`/`o` go — node jump and drill-down for any kind

caseycs:
## What problem this solves

Karpenter's `NodeClaim` is the object you land on when a node's provisioning goes wrong, but from a NodeClaim row there was no way across to the `Node`: `enter` opened the detail view, `o` refused (`'o' shows the node for a pod`), and `shift-j` follows the owner reference to the **NodePool**. Going the other way was just as manual: `enter` on a NodePool had nowhere to go, and reaching its NodeClaims meant `:nodeclaims` plus a hand-typed label filter.

Per review, this is not a Karpenter special case in control flow. Both jumps are one generic mechanism driven by data: a small built-in table that config can override or extend from `[views."…"]`.

## What this does

**`o` on any row that names a node** jumps to it, with the same breadcrumb and `esc` behaviour `o` already had for pods. Which field holds the name comes from a node-reference table, `NODE_REFS` in `src/views.rs`, that ships two rows — pods (`/spec/nodeName`) and NodeClaims (`/status/nodeName`) — so the jump works on first run for both. `enter` does the same for a kind without a built-in drill-down of its own.

```
nodeclaims ──enter/o──▶ nodes  [node of nodeclaim/default-sfpsl]
           ◀───esc────
```

**`[views."…"].node`** is a JSON Pointer that overrides a shipped row or adds a kind the table doesn't list:

```toml
[views."example.com/v1/widgets"]
node = "/status/host"
```

A row whose pointer is empty warns (`… has no node assigned`) rather than opening an empty list; a pointer that lands on a non-string warns that the pointer is wrong.

**`[views."…"].drill`** says where `enter` goes for a kind without a built-in drill-down: another kind, scoped by a label and/or field selector filled from the selected row (`{name}`, `{namespace}`):

```toml
[views."karpenter.sh/v1/nodepools"]
drill = { kind = "nodeclaims", labels = "karpenter.sh/nodepool={name}" }

[views.externalsecrets]   # the Secret it writes shares its name; nothing labels back
drill = { kind = "secrets", fields = "metadata.name={name}" }
```

`kind` is anything `:` accepts and is resolved when the key is pressed; an unknown kind warns and stays put. A namespaced target opens in the row's namespace, a cluster-scoped one ignores it. When a view sets both, `enter` drills and `o` still jumps to the node.

## User-visible changes to existing behaviour

- `o` is no longer pods-only. On any kind that names a node it jumps; on one that doesn't it warns `this kind names no node (see views."…".node)` instead of `'o' shows the node for a pod`.
- The breadcrumb for the pod jump moved from `host of web` to `node of pod/web`, so every jump to a node reads the same way and says which kind it came from.
- `enter` on a NodeClaim now jumps to its node instead of opening the detail view.

## Why `status.nodeName`

Karpenter pairs the two by provider ID (`AllNodesForNodeClaim` lists nodes with `spec.providerID == nodeClaim.status.providerID`), but `metadata.name` is the only field selector the apiserver indexes for nodes, so a `spec.providerID` watch selector isn't an option. `status.nodeName` is the same node's name, written onto the claim at registration and already the CRD's `Node` printer column. The `karpenter.sh/nodeclaim` label that v1beta1 synced onto nodes is gone in v1, so a label-selector approach wouldn't work either.

Refs: [`pkg/utils/nodeclaim/nodeclaim.go`](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/utils/nodeclaim/nodeclaim.go), [`pkg/controllers/nodeclaim/lifecycle/registration.go`](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/registration.go), [NodeClaims concepts](https://karpenter.sh/docs/concepts/nodeclaims/).

## Shape of the change

- `show_node` loses its kind check and reads the pointer from `views::node_pointer`; the node-scoping tail is `goto_node`, shared by `enter` and `o`.
- `drill_to_pods` becomes a thin wrapper over `drill_to(kind, …)`, the same tail resolved for an arbitrary kind. The `drill()` fallthrough consults `drill`, then `node`, then the detail view. Kinds with a built-in arm keep it.
- `node` and `drill` are resolved **key by key** through one `views::setting` helper, unlike `columns`/`sort` which come from the single most specific view — so a narrower view that only sets columns doesn't silently mask either. Documented in `docs/views.md`.
- `compile` validates both: a `node` that isn't a JSON Pointer, an empty `drill.kind`, or an unknown `{placeholder}` is dropped with a warning visible in `:config`.
- `Cluster::register_kind` lets a test declare the CRD it needs; `Cluster::fake()` is rebuilt on it, so the fixture is indexed the way `discover` indexes a real cluster.

## Tests and docs

- `src/app/tests.rs`: the pod `o` path (regression guard for the refactor), NodeClaim `enter`/`o` with no config plus `esc` unwind, the unregistered-claim warning, a kind that names no node, a bad pointer, a configured kind the table has never heard of, NodePool → NodeClaims, ExternalSecret → its Secret by field selector, a namespaced drill target named by alias, an unknown drill target, and drill-vs-node precedence.
- `src/views.rs`: the built-in table, config overriding a row and adding a kind, per-key resolution, pointer validation, drill compile and placeholder fill for `labels` and `fields`, drill validation.
- `docs/views.md` gains a "Navigating between kinds" section; the `enter`/`o` rows in `docs/keys.md` describe the generic behaviour.

`cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings` (with and without `--features bench`), and `cargo test` (535 passing) are clean.

### DIFF
--- a/docs/keys.md
+++ b/docs/keys.md
@@ -11,7 +11,7 @@ plugin, bookmark, and workspace chords.
 | `:<resource> <ns>`                            | switch kind and namespace at once (`:deploy social`; `all`/`*` = all namespaces; the namespace tab-completes)                         |
 | `[` / `]`                                     | view history - back / forward through visited kind+namespace views                                                                    |
 | `Tab` / `shift-Tab`                           | cycle views of the active workspace (when one is open)                                                                                |
-| `enter`                                       | drill down (workload/svc → pods, node → its pods, pod → containers, ns → re-scope, CRD → its resources)                               |
+| `enter`                                       | drill down (workload/svc → pods, node → its pods, pod → containers, ns → re-scope, CRD → its resources, kind with a `node` view → its node) |
 | `esc`                                         | go back / pop the view stack / clear filter / clear marks                                                                             |
 | `j`/`k`, `↓`/`↑`, `g`/`G`                     | navigate                                                                                                                              |
 | `S` / `I`                                     | sort-column picker (fuzzy; ⏎ on the active column inverts) / invert sort direction — remembered per kind across views and restarts    |
@@ -20,7 +20,7 @@ plugin, bookmark, and workspace chords.
 | `/`                                           | filter: fuzzy text · `!inverse` · `-l`/`-f` selectors (server-side on ⏎) · `status=X` `cpu>500m` `age<2h`                             |
 | `n` / `0`                                     | namespace switcher / all namespaces                                                                                                   |
 | `shift-j`                                     | jump to owner/controller                                                                                                              |
-| `o`                                           | show the node hosting the selected pod                                                                                                |
+| `o`                                           | show the node the selected row names (pods built in; other kinds via `[views."…"].node`)                                              |
 | `ctrl-r`                                      | refresh the watch                                                                                                                     |
 | `y` / `d` / `E`                               | view YAML / describe (`kubectl`) / live events                                                                                        |
 | `x`                                           | secrets: show `data` base64-decoded (as `stringData`)                                                                                 |
--- a/docs/keys.md
+++ b/docs/keys.md
@@ -11,7 +11,7 @@ plugin, bookmark, and workspace chords.
 | `:<resource> <ns>`                            | switch kind and namespace at once (`:deploy social`; `all`/`*` = all namespaces; the namespace tab-completes)                         |
 | `[` / `]`                                     | view history - back / forward through visited kind+namespace views                                                                    |
 | `Tab` / `shift-Tab`                           | cycle views of the active workspace (when one is open)                                                                                |
-| `enter`                                       | drill down (workload/svc → pods, node → its pods, pod → containers, ns → re-scope, CRD → its resources, NodeClaim → its node)         |
+| `enter`                                       | drill down (workload/svc → pods, node → its pods, pod → containers, ns → re-scope, CRD → its resources)                               |
 | `esc`                                         | go back / pop the view stack / clear filter / clear marks                                                                             |
 | `j`/`k`, `↓`/`↑`, `g`/`G`                     | navigate                                                                                                                              |
 | `S` / `I`                                     | sort-column picker (fuzzy; ⏎ on the active column inverts) / invert sort direction — remembered per kind across views and restarts    |
@@ -20,7 +20,7 @@ plugin, bookmark, and workspace chords.
 | `/`                                           | filter: fuzzy text · `!inverse` · `-l`/`-f` selectors (server-side on ⏎) · `status=X` `cpu>500m` `age<2h`                             |
 | `n` / `0`                                     | namespace switcher / all namespaces                                                                                                   |
 | `shift-j`                                     | jump to owner/controller                                                                                                              |
-| `o`                                           | show the node hosting the selected pod, or backing the selected Karpenter NodeClaim                                                   |
+| `o`                                           | show the node hosting the selected pod                                                                                                |
 | `ctrl-r`                                      | refresh the watch                                                                                                                     |
 | `y` / `d` / `E`                               | view YAML / describe (`kubectl`) / live events                                                                                        |
 | `x`                                           | secrets: show `data` base64-decoded (as `stringData`)                                                                                 |
--- a/docs/keys.md
+++ b/docs/keys.md
@@ -11,7 +11,7 @@ plugin, bookmark, and workspace chords.
 | `:<resource> <ns>`                            | switch kind and namespace at once (`:deploy social`; `all`/`*` = all namespaces; the namespace tab-completes)                         |
 | `[` / `]`                                     | view history - back / forward through visited kind+namespace views                                                                    |
 | `Tab` / `shift-Tab`                           | cycle views of the active workspace (when one is open)                                                                                |
-| `enter`                                       | drill down (workload/svc → pods, node → its pods, pod → containers, ns → re-scope, CRD → its resources, kind with a `node` view → its node) |
+| `enter`                                       | drill down (workload/svc → pods, node → its pods, pod → containers, ns → re-scope, CRD → its resources, kind with a `drill` view → that kind, kind with a `node` view → its node) |
 | `esc`                                         | go back / pop the view stack / clear filter / clear marks                                                                             |
 | `j`/`k`, `↓`/`↑`, `g`/`G`                     | navigate                                                                                                                              |
 | `S` / `I`                                     | sort-column picker (fuzzy; ⏎ on the active column inverts) / invert sort direction — remembered per kind across views and restarts    |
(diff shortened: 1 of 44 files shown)


nklmilojevic (changes requested):
This shouldn't land as a Karpenter special case.

Plugins can't do this jump - they're shell-out only, and they can't own `enter` or `o` anyway - so it does belong in core. The product bet though is one generic pipeline, not one arm per CRD. That's the k9s pattern we're deliberately not copying. Flux is the exception we called out as first-class. Karpenter isn't.

Keep the `goto_node` extract, drop the branded `"nodeclaims"` arms, and put the mapping in a small table (or in config next to `[views."…"]`). NodeClaim can be the first row / the example. The jump is just kind + JSON pointer → nodes.

Bare-plural matching is a bit loose but `helmreleases` already does that, so I'm not hung up on it. Tests and docs look good once it's generic.

nklmilojevic:
And thanks for the PR!

caseycs:
Thanks for the feedback. I did my best to implement the suggestions, make jumps nodeclaim -> node (also nodeclaim -> nodepool) and vise-versa generic and defined only via configuration.

I'm only familiar with basics of Rust, so I'm only Claude who did the coding. I tried to grill it as good as can to make it sharp and following same guidelines as the rest of the codebase.

src/views.rs
```diff
@@ -78,6 +78,104 @@ pub struct View {
...
+    }
+    unknown
+}
+
+/// Kinds whose objects name a node, and the JSON Pointer that holds the name.
+/// A row that names a node can jump to it (`o`, and `enter` where the kind has
+/// no drill-down of its own). Only pods are built in — every other kind,
+/// custom resources included, says where its node name lives via
+/// `[views."…"].node` rather than being enumerated here.
+const NODE_REFS: &[(&str, &str)] = &[("pods", "/spec/nodeName")];
```
nklmilojevic: One row here means a Karpenter user upgrading still has to hand-write TOML to get the jump this PR opens with. Add the nodeclaims row:

```rust
const NODE_REFS: &[(&str, &str)] =
    &[("pods", "/spec/nodeName"), ("nodeclaims", "/status/nodeName")];
```

This isn't me walking back the earlier comment. What I didn't want was a `nodeclaims` arm in `drill()`, control flow growing a branch per CRD. A row in a data table that config can override is the pattern we already use - `columns.rs:329` ships `FLUX_OBJECT_COLUMNS` for `kustomizations | helmreleases`. Karpenter is on most EKS clusters now, same treatment applies. The generic mechanism stays exactly as is, it just also works on first run.

src/views.rs
```diff
@@ -78,6 +78,104 @@ pub struct View {
...
+
+/// A per-kind setting, resolved key by key in [`lookup`]'s precedence rather
+/// than off whichever single view `lookup` picks: a more specific view that
+/// sets only columns mustn't mask a setting on a broader key, which would
+/// ignore it with no way to see why.
+fn setting<'a, T: ?Sized>(
+    views: &'a HashMap<String, View>,
+    ar: &ApiResource,
+    pick: impl Fn(&'a View) -> Option<&'a T>,
+) -> Option<&'a T> {
```
nklmilojevic, line 190: Optional. This resolves key by key while `columns` and `sort` resolve off the single most specific view, so there are two rules in one config file. I think you picked the right one and the doc explains it, so keep it - noting it because I'd rather we move `columns`/`sort` onto the same resolution later than keep explaining the split.

src/app/navigation.rs
```diff
@@ -40,10 +40,49 @@ impl App {
             // A Flux HelmRelease bridges into the same native inspector:
             // enter opens the history of the Helm release it manages.
             "helmreleases" => self.drill_into_helmrelease(&obj),
-            _ => self.open_detail(),
+            // Everything else is configuration: a `[views."…"].drill` opens
+            // another kind scoped to this row; failing that, anything that
+            // names a node (`[views."…"].node`) drills into it. Pods name one
+            // too, but they drill into containers above.
+            _ => {
+                if let Some(drill) = self.configured_drill() {
```
nklmilojevic, line 51: Optional. A `drill` configured on a kind that already has a built-in arm above (pods, services, CRDs, helmreleases) is silently ignored here. docs/views.md says so, but a compile-time warning in `compile()` would save someone a puzzled hour wondering why their stanza does nothing.

src/app/navigation.rs
```diff
@@ -198,25 +237,63 @@ impl App {
...
         self.flash_err = false;
         self.start_watch();
     }
 
+    /// Scope the nodes list to one node by name — the shared tail of every
+    /// jump to a node. The name is what we scope the watch by because
+    /// `metadata.name` is the only field selector the apiserver indexes for
+    /// nodes — a resource that pairs with its node by some other identifier
+    /// (Karpenter's `status.providerID`, say) can't be selected on that.
+    pub(super) fn goto_node(&mut self, node: &str, scope: String) {
```
nklmilojevic, line 284: Optional. `drill_to` sets `↳ drilled into {plural}` and this sets no flash, so `enter` gives different feedback depending on which path a kind takes. The scope label covers it in the header, so this is just evenness between two paths that feel like one action.

docs/keys.md
```diff
@@ -11,7 +11,7 @@ plugin, bookmark, and workspace chords.
 | `:<resource> <ns>`                            | switch kind and namespace at once (`:deploy social`; `all`/`*` = all namespaces; the namespace tab-completes)                         |
 | `[` / `]`                                     | view history - back / forward through visited kind+namespace views                                                                    |
 | `Tab` / `shift-Tab`                           | cycle views of the active workspace (when one is open)                                                                                |
-| `enter`                                       | drill down (workload/svc → pods, node → its pods, pod → containers, ns → re-scope, CRD → its resources)                               |
+| `enter`                                       | drill down (workload/svc → pods, node → its pods, pod → containers, ns → re-scope, CRD → its resources, kind with a `drill` view → that kind, kind with a `node` view → its node) |
```
nklmilojevic: This blew the table's alignment out, and the cell is now long enough to be hard to read. Shorten it and let views.md carry the detail.

nklmilojevic (changes requested):
Two things before merge. The first is inline on `NODE_REFS`: ship the nodeclaims row so the jump works without config.

The second is the description. It drifted while the branch evolved - it still says no docs changes, four tests, 522 passing, and describes the `nodeclaims` arm in `drill()` and the two-path `show_node`, none of which survived commit 3. Please also call out the two user-visible changes that came with the refactor: `o` now works on any kind that names a node rather than pods only, and the breadcrumb moved from `host of web` to `node of pod/web`. This body becomes the squash commit.

The rest of the inline comments are optional.

nklmilojevic:
> I'm only familiar with basics of Rust, so I'm only Claude who did the coding. I tried to grill it as good as can to make it sharp and following same guidelines as the rest of the codebase.

All good! :) You grilled Claude well, there just a few minor adjustments and it will be ready to go.

caseycs:
I've left two optional suggestions out of scope (2 still pending comments), but the rest is addressed.

PR #190, opened 2026-09-03
perf: cut per-frame work in the render, filter and log paths

vyrti:
## Summary

Implements the performance work no other open PR covers. Production code only —
no plan, no audit document, no prototype harness. Independent of #220–#228.

## What this changes, and by how much

Paired benchmarks: the shipped code and the code it replaced, in the same binary
and the same run, so the number is a ratio rather than a timing that only holds
at one thermal state.

| Change | Baseline | Now | Gain |
| --- | ---: | ---: | ---: |
| One clock reading per frame, not per elapsed cell — 2,000 AGE cells | 134.99 µs | 92.951 µs | **1.45x** |
| Log records without a JSON DOM — 10,000 records, 12 ingest fields | 19.441 ms | 8.096 ms | **2.40x** |
| Log records without a JSON DOM — 10,000 records, no extra fields | 5.563 ms | 3.451 ms | **1.61x** |
| Borrowed custom-column extraction — 100-element array, 2,000 objects | 1.757 ms | 796.40 µs | **2.21x** |
| Borrowed custom-column extraction — scalar | 669.36 µs | 547.67 µs | **1.22x** |
| Borrowed custom-column extraction — escaped annotation key | 532.61 µs | 448.64 µs | **1.19x** |
| SIMD base64 for the Helm payload's double decode | 353.60 ns | 195.75 ns | **1.81x** |

Dependency swaps:

| Change | Baseline | Now | Gain |
| --- | ---: | ---: | ---: |
| `foldhash` — look up every key in the store, 20,000 rows | 382.12 µs | 132.37 µs | **2.89x** |
| `foldhash` — same, 2,000 rows | 35.505 µs | 13.676 µs | **2.60x** |
| `nucleo` replaces the skim matcher — 20,000 names | 6.393 ms | 2.090 ms | **3.06x** |
| `compact_str` for table cells — 2,000 rows' worth | 213.82 µs | 119.90 µs | **1.78x** |

Memoized models, absolute timings — memo hits replacing a full rebuild on every
draw and again on every keystroke:

| Model | Now |
| --- | ---: |
| `display_headers` | 5.209 ns |
| `filtered_sort_entries` | 8.869 ns |
| `filtered_contexts`, 2,000 contexts | 8.108 µs |

Also: the gunzip buffer is pre-sized from the gzip ISIZE trailer, clamped so a
hostile Secret cannot turn a hint into a 4 GiB reservation; highlight positions
are memoized per name instead of rescored per visible row per frame; and a
typical pod row now allocates nothing, asserted by test.

## Dependencies

`+foldhash`, `+nucleo-matcher`, `+compact_str`, `−fuzzy-matcher`. The crate
graph shrinks by one: `foldhash` and `compact_str` were already compiled in
through `ratatui`, so only `nucleo-matcher` is new, and it displaces
`fuzzy-matcher` and its `thread_local`.

Case handling, match positions and filter syntax are held where they were; the
score scale changes, so equally-matching picker entries may order differently.
simd-json was tried and dropped — 1.05x, 0.97x and 0.94x over three runs on the
only large document sofka parses itself.

Environment: Apple M1, macOS/Darwin 25.6, AArch64. Criterion
`--warm-up-time 1 --measurement-time 2 --sample-size 20 --noplot`.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test` — 628 passed, 1 ignored (601 on `main`)

### DIFF
--- a/benches/hot_paths.rs
+++ b/benches/hot_paths.rs
@@ -75,10 +75,11 @@ fn cells(c: &mut Criterion) {
 
     let pods: Vec<_> = (0..256).map(bs::pod).collect();
     let pod_spec = columns::build_spec("pods", None, None, false);
+    let now = columns::now_secs();
     g.bench_function("pods_256", |b| {
         b.iter(|| {
             for o in &pods {
-                black_box(pod_spec.cells(o));
+                black_box(pod_spec.cells(o, now));
             }
         });
     });
@@ -89,7 +90,7 @@ fn cells(c: &mut Criterion) {
     g.bench_function("helm_16", |b| {
         b.iter(|| {
             for o in &helm {
-                black_box(helm_spec.cells(o));
+                black_box(helm_spec.cells(o, now));
             }
         });
     });
@@ -286,6 +287,27 @@ fn helm_decode(c: &mut Criterion) {
             black_box(v)
         });
     });
+    // The double base64 decode Helm's payload needs, SIMD against scalar.
+    {
+        let wire = secret
+            .data
+            .pointer("/data/release")
+            .and_then(serde_json::Value::as_str)
+            .expect("fixture payload")
+            .to_string();
+        let scalar = base64::engine::general_purpose::STANDARD;
+        let simd = base64::engine::Simd::standard(Default::default());
+        fn twice<E: base64::Engine>(e: &E, wire: &str) -> Vec<u8> {
+            let inner = e.decode(wire).expect("outer base64");
+            e.decode(inner).expect("inner base64")
+        }
+        g.bench_function("base64_scalar", |b| {
+            b.iter(|| black_box(twice(&scalar, black_box(&wire))));
+        });
+        g.bench_function("base64_simd", |b| {
+            b.iter(|| black_box(twice(&simd, black_box(&wire))));
+        });
+    }
     g.bench_function("parse_typed", |b| {
         b.iter(|| black_box(sofka::helm::parse_release_json(black_box(&json))));
     });
@@ -303,14 +325,47 @@ fn cell_extract(c: &mut Criterion) {
     g.bench_function("borrowed_ip_2000", |b| {
         b.iter(|| {
             for pod in &pods {
-                black_box(spec.cell_at(black_box(pod), 4).unwrap());
+                black_box(
+                    spec.cell_at(black_box(pod), 4, columns::now_secs())
+                        .unwrap(),
+                );
             }
         });
     });
     g.bench_function("owned_ip_baseline_2000", |b| {
         b.iter(|| {
             for pod in &pods {
-                black_box(spec.cell_at(black_box(pod), 4).unwrap().into_owned());
+                black_box(
+                    spec.cell_at(black_box(pod), 4, columns::now_secs())
+                        .unwrap()
+                        .into_owned(),
+                );
+            }
+        });
+    });
+    g.finish();
+}
+
+/// The per-frame clock. `now` measures the shipped path — one reading threaded
+/// through the row — against the previous one, where each elapsed cell read the
+/// clock for itself.
+fn frame_clock(c: &mut Criterion) {
+    let mut g = c.benchmark_group("frame_clock");
+    let pods: Vec<_> = (0..2_000).map(bs::pod).collect();
+    let spec = columns::build_spec("pods", None, None, false);
+
+    g.bench_function("one_reading_per_frame_2000", |b| {
+        b.iter(|| {
+            let now = columns::now_secs();
+            for pod in &pods {
+                black_box(spec.volatile(pod, "pods", 0, now));
+            }
+        });
+    });
+    g.bench_function("one_reading_per_cell_2000", |b| {
+        b.iter(|| {
+            for pod in &pods {
+                black_box(spec.volatile(pod, "pods", 0, columns::now_secs()));
             }
         });
     });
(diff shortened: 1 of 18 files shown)


vyrti:
@nklmilojevic how much PRs you want for such cases? 6? 12? plan 3 will be in parts by 10 performance updates per PR
this one is +2000 lines, I think its unmanageable

nklmilojevic (changes requested):
[P1] Update Cargo.lock for the dependency changes before merging.

Reviewed head: 526ce4f3330c320b02f6cb2727d458a4a055eae7.

Cargo.toml adds compact_str, foldhash, and nucleo-matcher and removes fuzzy-matcher, but the committed Cargo.lock is unchanged. On a fresh copy of this commit, `cargo metadata --locked --offline --format-version 1` fails with:

```text
error: cannot update the lock file .../Cargo.lock because --locked was passed to prevent this
```

The release workflow builds with `cargo build --release --locked`, and Nix packaging uses the committed Cargo.lock, so this blocks release/packaging despite the green PR CI. The ordinary CI commands allow Cargo to update the lockfile automatically.

Please regenerate Cargo.lock with Cargo and commit the resulting dependency changes, then verify a locked build succeeds.

Otherwise, `just check` passed locally after Cargo automatically updated the lockfile: formatting, Clippy, and 628 tests passed with 1 ignored. I found no other merge-blocking issues.

I'm creating a PR that will cach this on CI in the future.

vyrti:
Fixed in ec2d3de.

Reproduced your failure first on a clean checkout of 526ce4f — `cargo metadata --locked --offline --format-version 1` exits 101 with exactly the error you quoted. Regenerated with Cargo (no hand-editing) and it now exits 0.

What actually moved, since the diff is a little wider than the three added crates suggests:

- **`nucleo-matcher` 0.3.1 added**, **`fuzzy-matcher` removed** along with its `thread_local` dependency. This is the only real resolution change.
- **`compact_str` and `foldhash` add no packages.** Both were already resolved as transitive dependencies (ratatui and hashbrown respectively); the manifest only promotes them to direct, so they appear solely in the `sofka` dependency list.
- **Four `windows-sys` edges move `0.52.0` → `0.61.2`.** This one is worth flagging since it looks unrelated. It is a consolidation onto a version the tree already carried, not a new dependency — `ring` still holds `0.52.0`, and no package entry was added. I isolated it: applying *only* the `fuzzy-matcher` → `nucleo-matcher` swap to a clean `main` worktree reproduces this identical diff, so it falls out of dropping `fuzzy-matcher` rather than from anything in this branch. It is also identical whether resolved online or `--offline`, so it is not index freshness.

Verified with the locked commands: `cargo metadata --locked --offline` exit 0, `cargo clippy --locked --all-targets -- -D warnings` clean, `cargo test --locked` 628 passed / 1 ignored — matching the count from your local run. I did not run the release build itself.

Worth noting for sequencing: the `--locked` CI change you mentioned has since landed on `main` (2ef161c), so this branch's own CI now exercises the gap that let this through.

vyrti:
Rebased onto current `main` (30cd059) and resolved. Now a single commit, `3b7e81f`.

Two separate things had to be dealt with since the last push:

**1. The compile error — a semantic conflict, not a textual one.** This branch renamed `fuzzy_match` to `score` for the nucleo swap and converted its 7 call sites. `636be0c` ("feat: add external plugin packages") added an 8th caller on `main` using the old name. Different lines, so git merges cleanly and GitHub reported the PR mergeable, but the merged tree calls a method that no longer exists. Converted; `src/app/input.rs:666` now uses `.score(name, q)` like its neighbours.

**2. #204 landed, and it optimizes the same hot path incompatibly.** `columns.rs` conflicted in 27 places, all the same clash:

- `main` (via #204) makes `CellFn` return `Cow<'a, str>` — borrow from the object, own only derived values.
- this branch made it return `CompactString` — short cells inline, no heap under 24 bytes.

They target the same cost from opposite ends: `Cow` wins on cells that already exist as strings on the object (`name`, `podIP`, `nodeName`), `CompactString` wins on formatted ones (`3d4h`, `1/2`, restart counts) that `Cow` must own anyway.

**Resolved in favour of `main`'s `Cow`.** The `CompactString` cell layer is gone, along with `compact_str` as a direct dependency and the `typical_cells_do_not_allocate` test, which asserted inline storage this branch no longer has. Everything else here is independent of that choice and is kept: the nucleo fuzzy matcher, the memoized row/header/picker models, the selective-visitor provider log parsing, the helm SIMD base64 decode, `foldhash`, and the per-frame clock.

That last one needed care rather than a blanket "take main": the `now` threading is this branch's, and `views::render_cell` merged cleanly with the 3-argument signature, so `cell_at` keeps its `now` parameter while returning `main`'s `Cow`. `job_duration` and `time_since` likewise take the frame clock. Benchmarks from both sides are kept — `cell_extract` and `provider_selection` from #204 alongside `frame_clock`, `custom_columns`, `provider_records`, `fuzzy` and `frame_models` — minus the `CompactString`-vs-`String` comparison, which no longer measures anything shipped.

Also folded the lockfile into this commit rather than keeping it separate, per AGENTS.md: the lockfile ships with the manifest change.

Verified against the merge with current `main`, not just the branch tip — that is what would have caught the first problem:

- `cargo metadata --locked --offline` exit 0
- `cargo fmt --all -- --check` clean
- `cargo clippy --locked --all-targets -- -D warnings` clean, and again with `--features bench`
- `cargo test --locked` — 671 passed, 1 ignored

Happy to take the `CompactString` idea back up as a follow-up against the `Cow` design, where the owned arm holds a compact string and borrowable cells stay zero-copy — that would get both wins, but it is new design work rather than conflict resolution.

src/fuzzy.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub fn score(&self, haystack: &str, needle: &str) -> Option<i64> {
+        let mut inner = self.inner.borrow_mut();
+        inner.sync(needle);
+        let Inner {
+            matcher,
+            atom,
+            haystack: buf,
+            ..
+        } = &mut *inner;
+        atom.score(Utf32Str::new(haystack, buf), matcher)
```
nklmilojevic: [P2] Preserve matching for context names with combining accents

With the default Unicode segmentation in nucleo, a context name cannot always match itself. For a context named `pre\u{0301}prod`, enter the same name in the context filter. Here, `\u{0301}` means the U+0301 combining acute accent. The result list becomes empty. I reproduced this through `handle_key`. The previous matcher returns a match for both the full name and the prefix `pre\u{0301}`.

Correct the Unicode conversion or use character-based matching. Add a test through `handle_key` for this input.

benches/hot_paths.rs
```diff
@@ -303,14 +325,47 @@ fn cell_extract(c: &mut Criterion) {
...
+fn frame_clock(c: &mut Criterion) {
+    let mut g = c.benchmark_group("frame_clock");
+    let pods: Vec<_> = (0..2_000).map(bs::pod).collect();
+    let spec = columns::build_spec("pods", None, None, false);
+
+    g.bench_function("one_reading_per_frame_2000", |b| {
+        b.iter(|| {
+            let now = columns::now_secs();
+            for pod in &pods {
+                black_box(spec.volatile(pod, "pods", 0, now));
```
nklmilojevic: [P2] Measure the AGE column in both clock benchmarks

Column 0 in this pod view is NAME. Thus, this call and the baseline call on line 368 always return `None`. Neither call calculates nor formats an elapsed-time value. The reported 9.01x result measures clock-call overhead with an empty result, so it does not establish the gain for AGE-cell rendering.

Find the AGE column by its header and use its index in both benchmarks. Check that `volatile` returns a value. Then run the benchmarks again and update the PR results.

nklmilojevic (changes requested):
Please correct the two issues in the inline comments before merge.

Validation: `just check` passed on this commit. The test suite had 671 passes and one ignored test. An additional test through `handle_key` failed for a context name with a combining accent.

src/fuzzy.rs
vyrti: Confirmed and fixed in 43b86ae. Normalization turned out not to be involved — `Normalization::Never` fails identically, so the conversion was not the cause.

The asymmetry is between how the two sides are built. An atom's needle always comes from grapheme clusters, one `char` per cluster (`pattern.rs`, both branches of `Atom::new_inner` call `chars::graphemes`), so `pre\u{0301}prod` collapses to the seven chars `preprod`. `Utf32Str::new` collapses the haystack the same way — and then, when the collapsed result is pure ASCII, returns the *original* bytes instead of the collapsed chars:

```rust
buf.extend(crate::chars::graphemes(str));      // ['p','r','e','p','r','o','d']
if buf.iter().all(|c| c.is_ascii()) {
    return Utf32Str::Ascii(str.as_bytes());    // p r e CC 81 p r o d  — 9 "chars"
}
```

So the haystack keeps the accent's two UTF-8 continuation bytes as if they were characters. Printing it gives `"preÌ\u{81}prod"`, length 9, against a 7-char needle that no longer has them. Scoring the same atom against `Utf32Str::Unicode(&graphemes)` returns `Some(192)`, which pins it to that constructor rather than to the atom.

`score` and `indices` now build the haystack from the collapsed chars. The ASCII fast path is untouched: a name with no cluster to collapse returns on the `is_ascii()` branch above it, so ordinary names keep the byte-slice representation and cost nothing extra.

Test through `handle_key` as asked: `context_picker_matches_a_name_with_a_combining_accent` types the name a character at a time into the picker and asserts the list contains it, then backspaces to the `pre\u{0301}` prefix and asserts the same. It is verified to catch the bug — reverting just the two `utf32(...)` calls to `Utf32Str::new` fails it with `left: []`, the empty list you saw.

benches/hot_paths.rs
vyrti: Confirmed and fixed in 5d0f7e3. You are right on every point, and I verified each before changing anything: the pods view headers are `["NAME", "READY", "STATUS", "RESTARTS", "AGE"]`, so index 0 is NAME and AGE is index 4. For the bench fixture, `volatile(pod, "pods", 4, now)` returns `Some("7d")` while `volatile(pod, "pods", 0, now)` returns `None` — `volatile_cell` only answers for AGE and a couple of others, so both halves were timing a failed match.

`frame_clock` now looks the column up by header and asserts it renders, so a column that measures nothing fails the benchmark rather than reporting a number:

```rust
let age = spec.header_index("AGE").expect("pods view must have an AGE column");
assert!(
    spec.volatile(&pods[0], "pods", age, columns::now_secs()).is_some(),
    "AGE must render an elapsed value, or this measures nothing"
);
```

Re-measured over 2,000 pods, both halves now formatting a real elapsed value:

| | median |
| --- | ---: |
| `one_reading_per_cell_2000` | 134.99 µs |
| `one_reading_per_frame_2000` | 92.951 µs |

**1.45x**, not 9.01x. That is the honest shape of it: the old number was the ratio between two clock-call costs with no cell work on either side, so the clock was the entire measurement. Once AGE is actually formatted the clock is a much smaller share, and the gain shrinks accordingly. PR description updated with the corrected row.

nklmilojevic:
LGTM! Merging.

PR #194, opened 2026-09-04
perf: accept gzip-compressed API responses

vyrti:
**Offtopic:**
Interesting thing: this is only for legacy fallback that uses LISTS not new watch
as WATCH does not support GZIP in kubernetes, instead it seems to have protobuf
BUT

The kube-rs issue is kube-rs/kube#371 (https://github.com/kube-rs/kube/issues/371) — "protobuf encoding exploration", opened 
  Jan 2021 and now closed. Two structural blockers, per the maintainer:

  1. Kubernetes' protobuf field-ordering tags live in Go struct tags and aren't in the OpenAPI/Swagger schemas that k8s-openapi
     generates Rust types from — so the ordering metadata simply isn't available.
  2. Generated protobuf structs would collide with the existing k8s-openapi structs with no way to associate them. "Unless 
     there's a way to associate these structs with the ones from k8s-openapi structs of the same name, this would be hard. Sounds
     like another codegen project if it is possible."
     
https://github.com/kube-rs/kube/issues/725
**But it might worth review this as this was 5 years ago, as it brings benefits in gzip + zero copy deserialization for watch that rust ecosystem does not have now**
---
**PR Desc**
Enables kube's `gzip` feature. No source change: kube installs its decompression
layer inside `Client::try_from`, which `Cluster::from_config` already calls, so the
client starts advertising `Accept-Encoding: gzip` and inflating responses.

Measured against a local cluster — 771 pods, 1,222 configmaps, k8s 1.36 — reading
wire bytes straight off the apiserver. Times are transfer + inflate, median of 9:

| endpoint | plain | gzip | ratio | loopback |
| --- | ---: | ---: | ---: | ---: |
| `/api/v1/pods` | 3.88 MB | 202 KB | 19.2× | 75 → 45 ms |
| `/api/v1/configmaps` | 3.05 MB | 121 KB | 25.3× | 55 → 33 ms |
| `/api/v1/secrets` | 1.47 MB | 46 KB | 31.7× | 28 → 20 ms |
| `/api/v1/events` | 823 KB | 66 KB | 12.5× | 22 → 17 ms |
| `/apis/apps/v1/deployments` | 254 KB | 18 KB | 13.8× | 14 → 10 ms |

Loopback is the worst case for compression and it still wins by 1.7×. Projected
from the byte counts, the pods list takes 3,105 → 165 ms on a 10 Mbit link and
310 → 19 ms on 100 Mbit. Break-even is around 9 Gbit/s.

Cost: 0.7–3.3 ms to inflate (flate2 — 3.28 ms for the 3.8 MB pods list), +33 KB on
the stripped release binary (+0.23%), 3 transitive crates.

Two limits worth knowing:

- The apiserver only compresses responses over **128 KB**, so aggregated discovery
  (28 KB here), namespaces and nodes are untouched.
- **Watches are never compressed.** The main table view uses `streaming_lists()`
  (`src/k8s.rs:335`), so its initial sync — 3.93 MB here — saves nothing. The wins
  are `:find`, `:pulse`, the nodes view, the 5s metrics poll, events and `:fleet`.

`the_client_negotiates_and_inflates_gzip` covers both halves: the request carries
`accept-encoding: gzip`, and a gzip-encoded body is inflated before deserialization.
It fails without the feature flag.

`Cargo.lock` is included because `release.yaml` builds with `--locked`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,6 +155,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-compression"
+version = "0.4.43"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3976abdc8fe7d1133d43d304afd42abdf5bc3e1319d263d223bde07b5efc4be8"
+dependencies = [
+ "compression-codecs",
+ "compression-core",
+ "pin-project-lite",
+ "tokio",
+]
+
+[[package]]
 name = "async-stream"
 version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -438,6 +450,23 @@ dependencies = [
  "ryu",
  "static_assertions",
 ]
+
+[[package]]
+name = "compression-codecs"
+version = "0.4.38"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ce2548391e9c1929c21bf6aa2680af86fe4c1b33e6cea9ac1cfeec0bd11218cf"
+dependencies = [
+ "compression-core",
+ "flate2",
+ "memchr",
+]
+
+[[package]]
+name = "compression-core"
+version = "0.4.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc14f565cf027a105f7a44ccf9e5b424348421a1d8952a8fc9d499d313107789"
 
 [[package]]
 name = "convert_case"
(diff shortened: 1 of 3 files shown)


vyrti:
@nklmilojevic any concerns about this PR?
looks like clear win and used by k9s
conflicts/rebase will be fixed indeed

PR is only about gzip on fallback not on lege story about protobuf, but I must mention that anyway

nklmilojevic:
@vyrti will review more carefully on Monday - both perf updates!

nklmilojevic:
@vyrti ready to maerge after resolving conflicts

nklmilojevic (changes requested):
LGTM, we will merge after fixing the conflicts.

vyrti:
@nklmilojevic what to do with cases when PR fail check because there are no cargo.lock and we cant modify cargo.lock according to agents.md?

nklmilojevic:
The rule means don’t edit Cargo.lock manually. Regenerating it with Cargo when your change adds dependencies or enables features is expected - commit that update alongside Cargo.toml. Renovate owns routine dependency version bumps.

For this PR, enabling gzip adds transitive dependencies, so the lockfile update is correct and necessary for --locked checks.

vyrti:
@nklmilojevic ready for merge

PR #204, opened 2026-09-04
perf: reduce state and cell extraction overhead

vyrti:
Part 3 of 3 from the first optimization plan (I have a second one as well). Rebased onto current main (v0.23.0)

Three changes: curated cell extraction borrows instead of allocating, provider discovery drops a temporary vector, and UI state persistence moves off the render thread.

## What each change is actually worth

**`Cow<str>` cell extraction — the real hot-path win.**

Curated column extractors returned `String`. Filtering and sorting read *one* column of *every* object on every rebuild, so each keystroke in a structured filter allocated a `String` per row and dropped it immediately. They now return `Cow<str>`, borrowing values that already live in the Kubernetes object. Cached full rows call `into_owned()` and pay exactly what they paid before, so the full-row path is unchanged — the saving is confined to the one-column probes (`column_cell`, `sort_value`) that run per object per rebuild.

**State persistence off the render thread — a tail-latency fix, not a throughput one.**

Namespace, sort, and fleet changes used to TOML-encode and write inline in the input handler, on the render/event-loop thread. On a healthy filesystem that costs ~56µs, which nobody can perceive. The reason to move it is the tail: `create_dir_all` + `fs::write` on a stalled NFS mount, a fuse filesystem, or a synced home directory can block for seconds — and it was blocking *rendering and input*, so the whole TUI froze mid-keystroke. Writes now go to one ordered worker, which also coalesces a queued burst to the newest snapshot per destination file.

The table below reports 18x on UI latency. That number is real, but it measures the mean, which was never the problem. The claim worth reviewing is that a slow disk can no longer freeze the UI.

**Provider selection — marginal, kept for readability.**

`pick_service` / `pick_metrics_service` collected every candidate into a `Vec` before calling `min_by_key`; they now feed the filtered iterator straight in. Selection and port precedence are unchanged (`Iterator::min_by_key` returns the first minimum in both forms). This runs once per log-view or rightsize open and is dominated by a network round-trip, so the measured ~1.2x is worth nothing on its own — the code is just shorter.

## What going async cost

Moving the write off-thread relocated the stall rather than removing it. Most of the review so far has been about that, not about the optimization:

- **Teardown could hang.** `Drop` joined the worker unconditionally, so a stalled filesystem hung the process *after* the terminal had been restored — a quit that never returns, with nothing on screen to explain it. Teardown now waits on a channel the worker disconnects when it returns, which gives the wait a deadline that `JoinHandle::join` cannot have. Past 500ms the worker is detached and the process exits.

- **Failure notices could vanish.** The worker cannot write to stderr — it runs under the TUI's alternate screen, and ratatui only repaints cells it sees change, so a stray line stays smeared across the table. So notices go to the status line over the event channel. But a successful `try_send` is not proof of delivery: if the event loop stops before handling a queued message, the notice is lost and nothing records it. Failures are now recorded when they happen and cleared only when `App::handle_msg` acknowledges the notice by id; whatever is still pending at teardown is printed from `Drop`, which runs after the terminal is restored and so can safely use stderr. Ids are per occurrence, so an acknowledgement for one cannot clear a later, still-unhandled one — and the exit summary collapses them back to one line per distinct problem.

State-write failures also got their own diagnostics field. `:info` previously filed them under **Watch health**, next to the watch error count, where a disk permission error read as a broken watch.

## A latent bug found on the way: torn state files

Independent of everything above, and worth more than the microseconds: `save()` wrote in place. `File::create` truncates the target before the new TOML lands, so a crash, a power loss, or a second sofka writing the same file in that window leaves a half-written mix — and `load()` treats an unparsable file as an empty one, so the user's remembered sorts, namespaces, or fleet marks are silently gone. Moving the writes off the render thread widened that window; it did not open it. The inline version on main has the same hole.

The bytes now go to a sibling temp file that is flushed and renamed over the target. `rename` replaces atomically, so a reader sees the whole old file or the whole new one, never a tear. The temp file is fsynced, because otherwise the rename can reach disk before the bytes it publishes and a power loss leaves the target pointing at a zero-filled file — exactly the tear the rename is there to prevent. The directory entry is deliberately *not* synced: if the rename itself is lost, the previous complete file survives, which is a fine outcome for remembered UI state and saves a second flush per keystroke that changes a sort.

All three `save()` bodies were byte-identical, so this is one `atomicfile::write` rather than the same eight lines pasted three times. It sits in its own module rather than in `state_writer` so that `atomicfile::write` at a call site cannot be misread as enqueueing to the state-write worker. Four tests cover it: replace-in-place with parent creation, no temp litter after repeated writes, and a forced-failure case (renaming over a directory) asserting the old file survives untouched and the temp is cleaned up.

**Not converted:** `app/actions.rs`, `app/bundle.rs`, and `app/snapshot.rs` still use plain `fs::write`. Those are one-shot user-initiated exports to fresh paths, not state files rewritten under the app's own feet, so the tearing risk there is different in kind — and they are on the async side, so converting them needs a `spawn_blocking` wrapper or a duplicate async path. Happy to do it in a follow-up if you want the guarantee to be uniform.

## Benchmarks

| Benchmark | Baseline | Optimized | Result |
|---|---:|---:|---:|
| Cell extraction, 2,000 pod IPs | 167.23 µs | 111.31 µs | 1.50× faster |
| Log provider selection, 256 services | 4.688 µs | 4.029 µs | 1.16× faster |
| Metrics provider selection, 256 services | 4.867 µs | 4.074 µs | 1.19× faster |
| State UI latency, 200 writes/sample | 61.766 µs/write | 3.277 µs/write | 18.85× faster |
| State burst completion, 200 writes/sample | 61.766 µs/write | 3.901 µs/write | 15.83× faster |

`cargo bench --features bench --bench hot_paths -- "cell_extract|provider_selection"` and `cargo run --release --example state_write_probe --features bench`.

## Review notes

`state_writer.rs` is 291 lines of implementation plus 257 of tests, replacing three inline `save()` calls. That is a lot of machinery for a write that happens a few dozen times a session, and it is the part of this PR most worth pushing back on. The tail-latency argument is what justifies it — if you don't buy that, the `Cow` change stands on its own and this piece can be dropped without touching it.

617 tests passing, `clippy -D warnings` and `cargo fmt --check` clean. The behaviours that are easy to regress are pinned by tests that were checked against mutations — removing the teardown deadline, the drain redirect, or the summary dedup each fails its test rather than passing quietly.

### DIFF
--- a/src/atomicfile.rs
+++ b/src/atomicfile.rs
@@ -1,0 +1,137 @@
+//! Crash-safe replacement of the small state files sofka rewrites.
+//!
+//! Sort, namespace, and fleet choices are rewritten whole every time one
+//! changes. Writing them in place is not safe: `File::create` truncates the
+//! target before the new bytes land, so a crash, a power loss, or a second
+//! sofka writing the same file can leave a half-written mix that `load()`
+//! then silently discards as unparsable — the user's remembered choices gone
+//! because the process died at the wrong microsecond. Renaming a finished
+//! temp file over the target has no such window.
+
+use std::ffi::OsString;
+use std::io::Write as _;
+use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicU64, Ordering};
+
+/// Separates concurrent writes from this process; see [`temp_path`].
+static NEXT_TEMP: AtomicU64 = AtomicU64::new(0);
+
+/// Write `contents` to `path` as one indivisible replacement, creating the
+/// parent directory if it is missing.
+///
+/// The bytes go to a sibling temp file that is flushed and then renamed over
+/// the target. `rename` replaces atomically, so every reader — another sofka,
+/// a `cat`, this process on its next launch — sees either the whole previous
+/// file or the whole new one, never a tear.
+pub fn write(path: &Path, contents: &str) -> Result<(), String> {
+    if let Some(dir) = path.parent() {
+        std::fs::create_dir_all(dir).map_err(|e| e.to_string())?;
+    }
+    let temp = temp_path(path);
+    write_then_rename(&temp, path, contents).map_err(|e| {
+        // A failed attempt must not litter the state directory with a
+        // `sort.toml.tmp…` nobody will ever clean up.
+        let _ = std::fs::remove_file(&temp);
+        e.to_string()
+    })
+}
+
+fn write_then_rename(temp: &Path, path: &Path, contents: &str) -> std::io::Result<()> {
+    let mut file = std::fs::File::create(temp)?;
+    file.write_all(contents.as_bytes())?;
+    // Without this the rename can reach disk before the bytes it publishes,
+    // so a power loss leaves the target pointing at a zero-filled file —
+    // exactly the tear the rename is here to prevent. The directory entry is
+    // deliberately left unsynced: if the rename itself is lost, the previous
+    // complete file survives, which is a fine outcome for remembered UI state
+    // and saves a second flush on every keystroke that changes a sort.
+    file.sync_all()?;
+    drop(file);
+    std::fs::rename(temp, path)
+}
+
+/// A sibling of `path` — `rename` is only atomic within one filesystem, so the
+/// temp file cannot live in `/tmp` — that no concurrent writer can collide
+/// with: another sofka has another pid, and the counter separates writes
+/// racing inside this process.
+fn temp_path(path: &Path) -> PathBuf {
+    let n = NEXT_TEMP.fetch_add(1, Ordering::Relaxed);
+    let mut name = OsString::from(path.as_os_str());
+    name.push(format!(".tmp{}.{n}", std::process::id()));
+    PathBuf::from(name)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn scratch(tag: &str) -> PathBuf {
+        let dir = std::env::temp_dir().join(format!(
+            "sofka-atomicfile-{}-{tag}-{}",
+            std::process::id(),
+            NEXT_TEMP.load(Ordering::Relaxed)
+        ));
+        let _ = std::fs::remove_dir_all(&dir);
+        dir
+    }
+
+    #[test]
+    fn creates_missing_parents_and_replaces_in_place() {
+        let dir = scratch("replace");
+        let path = dir.join("nested").join("sort.toml");
+        write(&path, "first").unwrap();
+        assert_eq!(std::fs::read_to_string(&path).unwrap(), "first");
+        write(&path, "second").unwrap();
+        assert_eq!(std::fs::read_to_string(&path).unwrap(), "second");
+        let _ = std::fs::remove_dir_all(&dir);
+    }
+
+    #[test]
+    fn leaves_no_temp_file_behind() {
+        let dir = scratch("cleanup");
+        let path = dir.join("sort.toml");
+        write(&path, "a").unwrap();
+        write(&path, "bb").unwrap();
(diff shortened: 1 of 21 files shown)


vyrti:
@nklmilojevic not ready for merge, but ready for **careful** review now

src/atomicfile.rs
```diff
@@ -0,0 +1,137 @@
...
+    write_then_rename(&temp, path, contents).map_err(|e| {
+        // A failed attempt must not litter the state directory with a
+        // `sort.toml.tmp…` nobody will ever clean up.
+        let _ = std::fs::remove_file(&temp);
+        e.to_string()
+    })
+}
+
+fn write_then_rename(temp: &Path, path: &Path, contents: &str) -> std::io::Result<()> {
+    let mut file = std::fs::File::create(temp)?;
```
nklmilojevic: [P2] Create temporary files exclusively

Temp names use only the PID and a process-local counter. Two sofka instances on different hosts or in separate PID namespaces can have the same values while sharing a state directory (for example, an NFS home directory). `File::create` then opens and truncates the other writer's temporary file instead of detecting the collision. After one instance renames that inode into place, the other instance can continue writing through its existing handle and corrupt the published state file, defeating the atomic-replacement guarantee.

I reproduced this by keeping a colliding temporary file open, calling `atomicfile::write` successfully, then writing through the original handle: the published contents changed from valid TOML to an incomplete value.

Please use exclusive creation (`create_new(true)`) with collision retries, and only clean up temporary files this invocation successfully created. Add a regression test proving an existing temporary file is neither truncated nor removed.

nklmilojevic (changes requested):
One P2 correctness issue in atomic state-file replacement needs fixing; see the inline comment. No other actionable findings in the column, provider, or worker changes. Validation passed: formatting, Clippy including benchmark targets, and 617 tests (1 ignored); Cargo.lock remained unchanged.

src/atomicfile.rs
vyrti: Confirmed and fixed in 12b5eda.

I reproduced the truncation half standalone before changing anything — one handle writes 30 bytes to the temp name, a second `File::create` on the same name reports the file at **0 bytes**. And the collision is worse than truncation alone: in the test below, the second writer goes on to `rename` that shared inode into place, so the other instance's temp file *becomes* the published state file while its owner is still writing through the old handle.

`create_temp` now claims the name with `File::options().write(true).create_new(true)`, stepping to the next candidate on `AlreadyExists` and bounded at 16 attempts so a directory that refuses every create fails fast rather than spinning. Cleanup is then provably safe by construction rather than by inspection: a name that already existed was never opened, so the `remove_file` on the failure path cannot reach a file this call did not create.

Names now also mix in a clock reading alongside the pid and counter, but only to make the retry rare — `create_new` is what makes it correct. I did not reach for a random suffix, since that would mean a new dependency for a property the exclusive create already guarantees.

Two regression tests, both driving the real `write` path:

- `a_colliding_temp_file_is_neither_truncated_nor_removed` — the write is handed a name another writer holds, and must step over it: the squatter's bytes stay intact and the target still publishes.
- `a_failed_write_leaves_another_writers_temp_file_alone` — the same collision on the failure path, asserting the cleanup removes only the temp file this call created.

Both are verified to actually catch it: reverting just the `create_new` call back to `File::create` fails both (and passes the other four), with each panicking on reading back the squatter's file — which by then had been renamed away into the published state file.

Testing the collision needed a seam, since the real names mix in a clock reading and cannot be predicted from outside. `write` is unchanged in signature and behaviour; it delegates to a private `write_with` that takes the temp-name sequence as a closure.

PR #207, opened 2026-09-04
feat: port-forward picker with active-forward indicator column

steffen-karlsson:
- Replace the typed port-forward prompt with a single-select picker listing declared ports from the pod/service manifest, plus a "Custom…" entry that falls through to manual LOCAL:REMOTE input
- Add a dedicated headerless indicator column showing a teal ● next to rows with an active port-forward
- Extract port-label building into service_port_labels, pod_port_labels, and port_label helpers to flatten request_port_forward
- Update docs/keys.md, docs/features.md, and README.md

image

Screenshot 2026-09-04 at 22 26 16

### DIFF
--- a/src/app/actions.rs
+++ b/src/app/actions.rs
@@ -1210,6 +1210,7 @@ impl App {
         match cmd.spawn() {
             Ok(child) => {
                 let pf = PortForward {
+                    context: self.cluster.context.clone(),
                     cluster_url: self.cluster.cluster_url.clone(),
                     ns,
                     target,
@@ -1238,13 +1239,15 @@ impl App {
 
     /// Whether any live port-forward targets the given `(namespace, name)` on
     /// the current cluster. Used by the table renderer to mark forwarded rows.
-    /// Matched by cluster URL (not context name) so a kubeconfig reload that
-    /// remaps a context name to a different cluster won't show a stale marker.
+    /// Matched by both context name and cluster URL so neither a context name
+    /// remap nor same-server-different-credentials causes a false marker.
     /// The forward target may be prefixed with `svc/` for services.
     pub fn has_port_forward(&self, ns: &str, name: &str) -> bool {
+        let ctx = &self.cluster.context;
         let url = &self.cluster.cluster_url;
         self.port_forwards.iter().any(|pf| {
-            pf.cluster_url == *url
+            pf.context == *ctx
+                && pf.cluster_url == *url
                 && pf.ns == ns
                 && pf.target.strip_prefix("svc/").unwrap_or(&pf.target) == name
         })
@@ -2636,19 +2639,25 @@ fn service_port_labels(data: &Value) -> Vec<String> {
 
 /// Collect declared container ports from a Pod manifest as
 /// `"port:port  (container/portname)"` labels. Scans regular, init, and
-/// ephemeral containers so ports declared in any container type are offered.
+/// ephemeral containers. Init containers that have already terminated are
+/// skipped — their ports are no longer listening.
 fn pod_port_labels(data: &Value) -> Vec<String> {
     let mut out = Vec::new();
-    for path in [
-        "/spec/containers",
-        "/spec/initContainers",
-        "/spec/ephemeralContainers",
+    for (path, is_init) in [
+        ("/spec/containers", false),
+        ("/spec/initContainers", true),
+        ("/spec/ephemeralContainers", false),
     ] {
         let Some(containers) = data.pointer(path).and_then(Value::as_array) else {
             continue;
         };
         for c in containers {
             let cname = c.get("name").and_then(Value::as_str).unwrap_or("");
+            // Skip init containers that have already terminated — nothing is
+            // listening on their ports.
+            if is_init && container_terminated(data, "/status/initContainerStatuses", cname) {
+                continue;
+            }
             let Some(ports) = c.pointer("/ports").and_then(Value::as_array) else {
                 continue;
             };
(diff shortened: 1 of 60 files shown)


src/app/actions.rs
```diff
@@ -1234,6 +1228,22 @@ impl App {
...
+    /// the current cluster. Used by the table renderer to mark forwarded rows.
+    /// Matched by both context name and cluster URL so neither a context name
+    /// remap nor same-server-different-credentials causes a false marker.
+    /// The forward target may be prefixed with `svc/` for services.
+    pub fn has_port_forward(&self, ns: &str, name: &str) -> bool {
+        let ctx = &self.cluster.context;
+        let url = &self.cluster.cluster_url;
+        self.port_forwards.iter().any(|pf| {
+            pf.context == *ctx
+                && pf.cluster_url == *url
```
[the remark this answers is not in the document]
steffen-karlsson: Find this highly unlikely IMHO

steffen-karlsson:
I find the last comment highly unlikely, thus I closed it without a fix.

> **Context rename hides active forward**
> When the current kubeconfig context is renamed while a port-forward remains active, self.cluster.context changes but the forward retains its old context name, causing has_port_forward to return false and the active indicator to disappear.

nklmilojevic (changes requested):
Thanks for this! I like the port picker. It fits how sofka handles overlays and background actions.

I am requesting a few changes before we merge this:

- Please rebase on `main`. The PR currently has conflicts.
- Please keep the resource kind in the forward identity. At the moment, `svc/web` can mark any resource named `web`. Saved targets such as `pod/web` and `deploy/web` do not get a marker.
- Please avoid a separate hidden table column for the marker. It creates a second column index system. It also reserves space in every table. A marker in the existing `NAME` cell fits sofka better.
- Please filter UDP and SCTP ports. `kubectl port-forward` supports TCP only.
- Please add one test that starts with `handle_key('f')`. Please also add a render test for the marker.

The picker itself looks good. These changes will keep it aligned with sofka's generic table design. Thanks again!

steffen-karlsson:
> Thanks for this! I like the port picker. It fits how sofka handles overlays and background actions.
> 
> I am requesting a few changes before we merge this:
> 
> - Please rebase on `main`. The PR currently has conflicts.
> - Please keep the resource kind in the forward identity. At the moment, `svc/web` can mark any resource named `web`. Saved targets such as `pod/web` and `deploy/web` do not get a marker.
> - Please avoid a separate hidden table column for the marker. It creates a second column index system. It also reserves space in every table. A marker in the existing `NAME` cell fits sofka better.
> - Please filter UDP and SCTP ports. `kubectl port-forward` supports TCP only.
> - Please add one test that starts with `handle_key('f')`. Please also add a render test for the marker.
> 
> The picker itself looks good. These changes will keep it aligned with sofka's generic table design. Thanks again!
> 

Happy you like the port picker 🙏

Makes sense, I actually first did the dot inline with the names but found it a bit weird for my eyes, but I understand the empty column. Let me see if I can come up with anything else, otherwise I'll prepend the name.

I'll cover the rest and some extra tests tonight, stay tuned.

steffen-karlsson:
All good now @nklmilojevic 👍

src/ui.rs
```diff
@@ -1067,17 +1070,25 @@ fn all_ready(ready: &str) -> bool {
 /// fuzzy row filter (bold yellow) so a scan across many filtered results is
 /// faster — every visible row already matched, this just shows *where*.
 /// Falls back to a flat `base`-colored cell when there's no active filter.
-fn render_name_cell(app: &App, name: &str, base: Color) -> Cell<'static> {
+fn render_name_cell(app: &App, name: &str, base: Color, forwarded: bool) -> Cell<'static> {
+    // A teal ● prepended when a port-forward is active for this row.
+    let marker = if forwarded {
```
nklmilojevic, line 1184: [P2] Reserve layout space for the active-forward indicator

The marker adds two display columns to NAME, but the needed-width calculation still measures only the original name. With a LoadBalancer service named alpha and a long external hostname, I reproduced alpha rendering as `● alp` at terminal widths of 80 and 90 columns. Allocate a separate indicator column, as described in the PR, or include the prefix in width calculations so starting a forward does not unnecessarily clip the resource name.

src/ui.rs
```diff
@@ -752,7 +753,9 @@ fn draw_table(frame: &mut Frame, app: &mut App, area: Rect) {
                         c.into_cell_aligned(align)
                             .style(Style::default().fg(status_badge))
                     } else if i == name_col {
-                        render_name_cell(app, c.as_str(), row_color)
+                        let ns = obj.metadata.namespace.as_deref().unwrap_or_default();
+                        let fwd = app.has_port_forward(ns, c.as_str(), &app.kind_plural);
+                        render_name_cell(app, c.as_str(), row_color, fwd)
```
nklmilojevic: [P2] Preserve the active-forward indicator on marked rows

The marked_row branch above bypasses render_name_cell entirely. Starting a forward and then pressing Space on that row makes its `●` disappear even though the forward remains active. I reproduced this with the TestBackend. Render the indicator independently of the marked-row styling so bulk selection does not hide active forwards.

src/app/tests.rs
```diff
@@ -1519,7 +1532,408 @@ async fn port_forward_prompt_prefills_first_exposed_port() {
...
+        &mut app,
+        json!({
+            "apiVersion": "v1", "kind": "Service",
+            "metadata": {"name": "web", "namespace": "default", "resourceVersion": "1"},
+            "spec": {"ports": [{"port": 8080}]}
+        }),
+    );
+    app.table_state.select(Some(0));
+    app.request_port_forward();
+    app.handle_key(press(KeyCode::Enter)).unwrap();
```
nklmilojevic, line 1801: [P2] Stub subprocess launching in the forward-start tests

Confirming the picker here launches the real kubectl executable through start_port_forward. These new tests therefore depend on kubectl being installed, despite using Cluster::fake(). Running this test with PATH=/usr/bin:/bin fails at the next assertion with `port-forward failed to start: No such file or directory (os error 2)`. The pod-target and marker-render tests have the same dependency. Use a controlled subprocess stub or an injectable launcher so the unit suite works without Kubernetes tooling and does not invoke a real kubeconfig context.

nklmilojevic (changes requested):
Please address the three inline findings. Baseline just check passed with 618 tests passing and one ignored. Focused checks reproduced name clipping, loss of the active-forward marker on marked rows, and a new test failure when kubectl is absent from PATH.

PR #213, opened 2026-09-05
fix: flake binary

dminca:
* the better way of reaching out to binaries from the store instead of hardcoding any paths


nklmilojevic:
thanks for this!

dminca:
welcome, cheers :clinking_glasses:

PR #220, opened 2026-09-05
perf: plan 3: part 1

vyrti:
_Last updated: 2026-09-06 19:24 (GMT+1) — all numbers below re-measured against `main` @ 972f01f._

**OpenAI Astra 6** in action :P plan 3, around 90+ issues

Performance plan 3, part 1 — findings 001–010, verified against the code and implemented.

Fixed: cell-cache hits no longer allocate (001), filter masks and comparison
columns are resolved once per rebuild instead of per object, reading columns
from the per-revision cell cache (006), the renderer carries the store's own
row key instead of reformatting it (007), scrolled-away columns are no longer
formatted (009), mouse hit-testing derives column ranges instead of re-solving
the layout (010), plus buffer reuse in the row rebuild (005), borrowed header
labels and fit-first hints (008), a bounded message drain (002), tick redraws
only when something changed (003), and readiness-based headless snapshots (004).

`main` has moved on since this branch was cut, so everything below was
re-measured from scratch against current `main` (972f01f), which now includes
the column-stability fix (#243). That fix measures column widths over the whole
filtered list rather than the visible rows, and this branch defers to it — so
finding 009 now covers formatting only, not measurement.

## Wall clock (criterion, `main` → this branch)

Both trees measured back to back on an idle machine, 10s per benchmark.

| benchmark | main | branch | change |
|---|---|---|---|
| `filter_cmp/num/2000` | 2.501 ms | 440.8 µs | −82% |
| `filter_cmp/str_curated/2000` | 2.602 ms | 542.5 µs | −79% |
| `viewport/warm_47/500` | 6.79 µs | 1.52 µs | −78% |
| `viewport/warm_47/2000` | 6.51 µs | 1.48 µs | −77% |
| `filter/no_match/2000` | 155.1 µs | 106.3 µs | −31% |
| `helm_rows/refilter/300` | 24.92 µs | 18.63 µs | −25% |
| `filter/name_hit/2000` | 367.2 µs | 280.8 µs | −24% |
| `helm_rows/refilter/1200` | 24.86 µs | 19.71 µs | −21% |
| `filter/broad/2000` | 216.2 µs | 185.2 µs | −14% |
| `filter_cmp/str_ns/2000` | 82.14 µs | 76.48 µs | −7% |
| `rows_cache/event_then_rebuild/2000` | 9.33 µs | 9.37 µs | flat |
| `rows_cache/event_then_rebuild/500` | 8.56 µs | 8.66 µs | flat |

`rows_cache` is flat because an unfiltered, unsorted update never rebuilds —
`ensure_rows_cache` returns at its dirty guard, so none of this code runs there.

## Allocations (deterministic A/B, `main` → this branch)

Frame wall-clock has a wide run-to-run noise floor on the measuring machine, so
the render findings are measured by allocation count instead, which is
deterministic.

| probe | main | branch | change |
|---|---|---|---|
| `rebuild/cmp_status/2000` | 40,523 | 245 | −99.4% |
| `rebuild/cmp_restarts/2000` | 40,523 | 245 | −99.4% |
| `rebuild/fuzzy_miss/2000` | 2,243 | 244 | −89% |
| `rebuild/fuzzy_hit/2000` | 2,226 | 243 | −89% |
| `frame/event_then_frame/2000` | 6,024 | 3,731 | −38% |
| `frame/table_scrolled/2000` | 1,497 | 1,215 | −19% |
| `frame/table/500` | 1,782 | 1,488 | −16% |
| `rebuild/unfiltered/2000` | 212.8 | 212.8 | unchanged |

No probe regresses. Nothing in the branch was kept for its own sake: every
change above either moves a number or is required by one that does.

## Method

`main` has no `examples/allocprobe.rs` and no frame/viewport benchmarks, so the
baseline was taken in a worktree at 972f01f with this branch's harness copied
in — the probe, the benchmarks, `benchsupport`, and the `bench_install_kind`
fixture helper — and `warm_viewport` adapted to main's own viewport API. Only
measurement scaffolding was added; no production code was changed in the
baseline tree.

## Notes

- The pods benchmark fixture never installed the resolved kind or view spec, so
  App-based benchmarks were measuring `DEFAULT_COLUMNS`. Fixed here, and ported
  into the baseline tree before any number was taken, so both sides render the
  real pod columns.
- Comparison filters now warm the full cell cache for every object they test
  (previously one column per object): far cheaper in steady state, more work on
  the first rebuild after a view change. Elapsed-time columns stay live.
- `frame/event_then_frame` still carries main's width invalidation: a watch
  event drops the cached column widths, so the next frame re-measures every row
  in the list. That is what keeps this probe at 3,731 rather than the ~1,700
  this branch saw before #243 existed. Making that invalidation incremental is
  worth its own change; widening it here would trade the stable columns back.
- Not done: incremental row membership (005), no-op key redraws (003), caching
  the header/column layout by view revision (008).


src/main.rs
```diff
@@ -484,6 +505,24 @@ async fn snapshot(app: &mut App, rx: &mut mpsc::Receiver<store::Msg>) -> Result<
...
+/// initial list, the metrics the CPU/MEM columns show (pods/nodes only), and
+/// the dashboard `PUP_DEMO` asked for. Deliberately narrow — anything not
+/// rendered by the frame must not hold the snapshot open.
+fn snapshot_ready(app: &App) -> bool {
+    if !app.store.synced {
+        return false;
+    }
+    if app.metrics_columns() && app.metrics.is_empty() {
+        return false;
+    }
```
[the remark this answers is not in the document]
vyrti: Fixed in 0fe97b8, though the check was wrong in the other direction too.

Row coverage isn't something this window can promise either way: the metrics poll sleeps 5s between snapshots (`lifecycle.rs::spawn_metrics_poll`) and the deadline is 3s, so a second snapshot can never arrive inside it. The map present at t=0.5s is the same map at t=3s — holding the window open doesn't add the missing rows, and an object metrics-server hasn't reported renders here exactly as the live table renders it at that moment.

What was genuinely wrong is that `!metrics.is_empty()` used the poll's *result* to infer whether the poll had *reported*, which got two cases backwards: a namespace with no usage data yet reports an empty map, and a broken metrics API reports `Msg::MetricsError` and never fills the map. Both were treated as "not ready" and burned the full three seconds waiting for data that was never coming — the opposite of what the change was for.

Readiness now watches for `Msg::Metrics`/`Msg::MetricsError` going past in the loop, so it means "the poll has reported", empty or failed included, and no longer claims anything about coverage.

vyrti:
@nklmilojevic ready for review

nklmilojevic:
Lets fix the conflicts and merge in main so CI can run green and I'll review it afterwards.

src/main.rs
```diff
@@ -484,6 +517,30 @@ async fn snapshot(app: &mut App, rx: &mut mpsc::Receiver<store::Msg>) -> Result<
...
+fn snapshot_ready(app: &App, metrics_reported: bool) -> bool {
+    if !app.store.synced {
+        return false;
+    }
+    // Waiting for the poll to *report* is the most this can promise. It cannot
+    // promise a metrics row for every object: the poll runs every 5s, so a
+    // second snapshot can never arrive inside this 3s window, and an object
+    // metrics-server has not reported yet renders here exactly as the live
+    // table renders it at the same moment.
+    if app.metrics_columns() && !metrics_reported {
```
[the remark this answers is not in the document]
vyrti: Confirmed and fixed in 920c852.

The race is not just possible, it is the common case. `spawn_node_pods_poll` builds its publication ticker with `tokio::time::interval(Duration::from_secs(1))`, whose first tick fires immediately — at which point `synced` is still false, so nothing is published and the earliest possible `Msg::NodePods` is t≈1s. The quiet window is 250ms. Any cluster that syncs and reports metrics inside ~750ms closes collection before the counts exist, and `node_pods_cell` renders `-` for `None`. The fixed three-second wait had been covering for that.

`snapshot_ready` now also waits on `app.node_capacity_columns() && app.node_pods.is_none()` — the same predicate that gates pushing the PODS cell in `rows.rs` and `ui.rs`, so readiness and rendering cannot disagree about whether the column is shown.

No `node_pods_reported` flag was needed, unlike `metrics_reported`. That flag exists because an empty metrics map is ambiguous between "no usage data" and "not reported yet". `node_pods` is `Option<HashMap<..>>` and is set unconditionally on a matching-generation `Msg::NodePods`, so `is_some()` already means "has reported" and is the very `None` the cell renders from.

One accepted consequence: if the pods kind cannot be resolved, or both watch and list fallback are refused, a `--snapshot nodes` now waits out the full three-second deadline instead of finishing early. That is the pre-PR behaviour, it stays bounded by the deadline, and there is no `NodePodsError` message to distinguish "will never report" from "not yet" the way `MetricsError` allows for metrics.

vyrti:
@nklmilojevic done

nklmilojevic:
@vyrti wdyt of using stacked PRs for these perf plans? They get outdated pretty fast and require constant conflicts resolving - https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

vyrti:
@nklmilojevic they just touch many files in one PR
and I think they already stacked, so 220,221,222 can be merged in a row
solving latest conflicts now

vyrti:
@this one clean now again

src/ui.rs
```diff
@@ -908,30 +950,35 @@ fn draw_table(frame: &mut Frame, app: &mut App, area: Rect) {
...
+        let mut ranges = Vec::with_capacity(col_widths.len());
+        for (w, i) in col_widths
+            .iter()
+            .copied()
+            .zip((0..headers.len()).filter(|&i| col_visible(i)))
+        {
+            let start = x.min(cols_end);
+            let end = start.saturating_add(w).min(cols_end);
+            ranges.push((start, end, i));
+            x = end.saturating_add(2); // Table::column_spacing
```
nklmilojevic: [P2] Use the rendered column widths for mouse clicks

At a terminal width of 50 columns, clicking the visible AGE header selects RESTARTS. The fixed and capped widths can exceed the available space. Ratatui then reduces the column widths, but these mouse ranges still use the original widths. Clamping the end positions does not reproduce that layout.

Use the same layout calculation for rendering and mouse clicks. Add a test that clicks a header at its rendered position in a narrow terminal. The reproduction test passes on the base commit and fails here.

src/columns.rs
```diff
@@ -500,6 +500,24 @@ impl ViewSpec {
...
+    /// once per view whether a cached cell may be reused.
+    pub fn volatile_column(&self, plural: &str, idx: usize) -> bool {
+        let Some(col) = self.columns.get(idx) else {
+            return false;
+        };
+        match &col.source {
+            SpecSource::User(uc) => uc.kind == crate::views::ColumnKind::Time,
+            SpecSource::Curated(_) => matches!(
+                (plural, col.header.as_str()),
+                (_, "AGE") | ("jobs", "DURATION") | ("cronjobs", "LAST-SCHEDULE")
```
nklmilojevic: [P2] Treat Helm UPDATED columns as live values

This list omits UPDATED for both `helm` and `helmhistory`. The `col_helm_updated` extractor calculates elapsed time, but the new comparison filter reads its cached value. For example, cache a release at `2s`, let time advance, then enter `updated>2`. The filter still excludes the release because it compares against the old value.

Include these UPDATED columns in the live-value classification. Add a test through `handle_key` that applies the filter after time advances without a new Secret revision. The reproduction test passes on the base commit and fails here.

src/main.rs
```diff
@@ -517,6 +550,41 @@ async fn snapshot(app: &mut App, rx: &mut mpsc::Receiver<store::Msg>) -> Result<
...
+    // `node_pods` is `None` until that first publication, so the field already
+    // is the "has reported" state, and it is the same `None` the cell renders
+    // from.
+    if app.node_capacity_columns() && app.node_pods.is_none() {
+        return false;
+    }
+    match app.mode {
+        app::Mode::Pulse => app.pulse.nodes_total + app.pulse.pods_total > 0,
+        app::Mode::Xray => !app.xray_items.is_empty(),
+        _ => true,
```
nklmilojevic, line 622: [P2] Wait for custom columns before declaring a snapshot ready

A synced custom-resource list can reach this branch while its CRD printer-column request is still pending. The snapshot can then finish after the 250 ms quiet period with only NAME and AGE. In a reproduction, printer columns arrive after 500 ms. The base commit includes them, but this commit exits before they arrive. This can also omit a sort that depends on a printer column.

Track whether the printer-column request has finished. Wait for success or failure before declaring the snapshot ready, subject to the existing three-second deadline.

nklmilojevic (changes requested):
Please correct the three defects in the inline comments before merge.

Validation: `just check` passed on this commit, with 652 tests passed and one ignored. All three additional reproduction tests fail on this commit and pass on the base commit `b086f86e8ec0221c0f1ce8a4d063c6bf440a3ef4`.

src/main.rs
vyrti: Confirmed and fixed in 4bb2eed.

`snapshot_ready` now waits on a new `App::printer_columns_pending`, bounded by the existing three-second deadline. It is cleared at the top of `maybe_fetch_printer_columns`, set to the plural when a fetch actually spawns, and cleared again when the result lands.

One thing had to change before "wait for success **or** failure" was achievable as written: the fetch was dropping its failures.

```rust
let Ok(crd) = api.get(&name).await else { return };   // no message ever sent
```

A missing CRD, an aggregated API, or a refused permission sent nothing at all, so there was no "finished" signal to wait on — waiting on that state would have stalled every such snapshot for the full three seconds instead of ending after one round trip. It now reports either way:

```rust
let view = match api.get(&name).await {
    Ok(crd) => crate::views::printer_columns_view(&crd.data, &version),
    Err(_) => None,
};
```

That has a deliberate side effect worth calling out: an unreadable CRD is now cached as "nothing usable" for the session rather than re-fetched on every switch back to that kind. It matches what the per-plural cache's own doc comment already claimed it did, but it does mean a permission granted mid-session is not picked up until the next launch. Say the word if you would rather keep retrying, and I will track the pending request separately from the cache instead.

`printer_column_result_clears_the_pending_request` covers the clearing path, including the "nothing usable" shape a refused CRD now reports.

Worth flagging on coverage: `snapshot_ready` itself has no test, and this fix does not change that. It is a private fn in `main.rs`, and `Cluster::fake()` is `#[cfg(any(test, feature = "bench"))]` — the binary target links the library built without `cfg(test)`, so a test there does not compile under plain `cargo test`. The same gap applies to the node-pods readiness fix and to 83aa65f earlier in this PR. Making it testable means lifting the predicate into the library as an `App` method; happy to do that here if you want the coverage rather than a follow-up.

src/ui.rs
vyrti: Confirmed and fixed in f14724a.

You are right about the cause, and the assumption that produced it was written into the comment this PR added:

> The widths are already fixed `Length`s that fit the budget […] Running the constraint solver a second time […] was pure duplicate work.

They do not always fit. `distribute_column_widths` sizes `Exact` and `Cap` columns from their content, not from the budget — the `fixed` sum is computed first and `left` merely saturates to zero — so on a narrow terminal the declared total exceeds `content_budget`. The Table's solver then shrinks them, and a running sum over the declared widths puts every column after the first at an x that was never rendered. As you say, clamping the ends cannot reproduce that: the space is redistributed, not truncated.

Restored the shared layout, which is what the base commit did:

```rust
let rects = Layout::horizontal(widths.iter().copied())
    .flex(Flex::Start)
    .spacing(2)
    .split(cols_area);
```

Same widths, same spacing, same default `Start` flex the Table uses, so the recorded ranges are the geometry that was drawn. That is a second constraint solve per frame — the cost the running sum was trying to avoid — but the sum was only ever correct when nothing had to shrink.

Test as asked: `header_click_in_a_narrow_terminal_sorts_the_column_it_hits` renders into a 50-column `TestBackend`, scans the header row of the rendered buffer for where `AGE` actually appears, clicks that x, and asserts `sort_column` is AGE. It reads the position out of the buffer rather than out of `table_hit.cols`, so it cannot pass by agreeing with the geometry it is meant to be checking.

Verified to catch it: reverting just this hunk to the running sum fails the test with `left: None` — at that x the click lands outside every recorded range.

src/columns.rs
vyrti: Confirmed and fixed in b4ecdbd.

`col_helm_updated` computes elapsed time from the clock:

```rust
Some(secs) => humanize((Timestamp::now().as_second() - secs).max(0)),
```

but `volatile_column` listed only `AGE`, `jobs`/`DURATION` and `cronjobs`/`LAST-SCHEDULE`, so `resolve_cmp_column` classified UPDATED as `CmpColumn::Cached(i)` and the comparison read the warmed cell instead of re-running the extractor. Both helm plurals are now on that list.

Worth noting why the classification is deliberately wider than `volatile_cell`, since the two now disagree: `volatile_cell` is the cheap per-frame path and has no way to produce this value — it takes the raw object, while UPDATED needs the decoded release summary (base64 + gunzip + JSON). The filter reaches it through `CmpColumn::Live(i) => self.spec.cell_at(o, i, now)`, which re-runs the real extractor. So UPDATED is live for the filter but still renders from the cache in the table, which is the pre-existing behaviour and not something this PR changed. I left it there rather than widen the scope; happy to make the rendered cell refresh too if you want that, though it would mean a helm decode per visible row per frame.

Test through `handle_key` as asked: `helm_updated_filter_reads_the_live_elapsed_value` applies a release deployed at `now`, types `updated>1` (which warms the cell cache and correctly excludes it at that instant), lets ~2.1s pass, then clears with Esc and retypes the same filter. Esc clears `self.filter` and invalidates the row order but does not drop `cache.cells`, so the cached cell survives — the Secret never changes revision, exactly the shape you described.

Verified to catch it: reverting just the two match arms fails the test with `left: []` — the release stays filtered out of its own list.

vyrti:
@nklmilojevic this one is ready for **careful** review

vyrti:
@nklmilojevic no conflicts, PR compared to current main, still a lot of win

src/main.rs
```diff
@@ -428,12 +433,40 @@ async fn snapshot(app: &mut App, rx: &mut mpsc::Receiver<store::Msg>) -> Result<
...
+        let now = tokio::time::Instant::now();
+        if now >= deadline {
+            break;
+        }
+        match tokio::time::timeout(SNAPSHOT_QUIET.min(deadline - now), rx.recv()).await {
+            Ok(Some(msg)) => {
+                metrics_reported |= matches!(
+                    msg,
+                    store::Msg::Metrics { .. } | store::Msg::MetricsError { .. }
+                );
```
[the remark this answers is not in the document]
vyrti: Confirmed and fixed in 2dc53c6. Snapshot readiness now counts `Metrics` and `MetricsError` only when the message generation matches `app.generation`, so a report from a superseded watch is discarded for readiness exactly as `handle_msg` discards it for application state. Added a regression test covering current and stale generations for both success and error reports. `just check` passes.

PR #221, opened 2026-09-05
perf: plan 3: part 2

vyrti:
> **Stacked on #220.** Both parts need the same benchmark harness, and two
> branches cut from `main` cannot each introduce it without colliding, so this
> branch is based on #220 rather than on `main`. GitHub shows #220's commits
> here until it merges; afterwards this diff is part 2 alone. Merge #220 first.

Performance plan 3, part 2 — findings 011–020, verified against the code.
Nine are real; 019's evidence is half out of date (its sequential selector
sweep is real, but the "candidate helpers sort" it describes was already
replaced by `min_by_key`).

Fixed: provider framing no longer rescans its buffer for a newline on every
chunk and hands parsers borrowed lines (017, 018); a huge log line is wrapped
only for the rows the viewport shows (012); log severity is detected in one
case-insensitive pass instead of a lowercased copy plus seventeen scans, and
the search needle is compiled once per frame rather than per styled span
(013, 016); NAME cells slice their highlight runs and borrow unfiltered names
(015); the namespace switcher stops cloning every name twice (011, partial);
service discovery drops the Vec it built only to take a minimum (019, partial).

## Wall clock

Measured **interleaved** (before/after alternating, two rounds), because
sequential runs on this machine carry a first-run/second-run bias of several
percent that reads as a change.

| benchmark | before | after | change |
|---|---|---|---|
| `provider_ingest/fragmented_512k` | 11.74 / 11.57 ms | 82 / 74 µs | **−99.3%** (~150×) |
| `log_frame/huge_line` | 1.02 ms / 977 µs | 869 / 878 µs | **−12.5%** |
| `provider_ingest/chunk/256` | 353 / 313 µs | 309 / 286 µs | **−10%** |
| `provider_ingest/chunk/2000` | 2.64 / 2.54 ms | 2.40 / 2.28 ms | **−9%** |
| `log_frame/plain` | 351 / 353 µs | 324 / 329 µs | **−7%** |
| `overlay_frame/help` | 290 / 293 µs | 294 / 292 µs | unchanged (control — path untouched) |

## Allocations per frame / per ingest

Deterministic A/B: a worktree at the base commit with an identical harness,
against this branch.

| probe | before | after | change |
|---|---|---|---|
| `log_frame/huge_line` | 4,757 | 382 | **−92%** (bytes −47%) |
| `doc_frame/filtered` | 586 | 392 | −33% |
| `log_frame/plain` | 717 | 495 | −31% |
| `log_frame/filtered` | 709 | 514 | −28% |
| `log_frame/wrapped` | 1,031 | 809 | −22% |
| `overlay/ns_browse` | 2,738 | 2,238 | −18% |
| `name_cells/filtered_frame` | 2,060 | 1,794 | −13% |
| `provider/chunk_2000` | 50,015 | 48,004 | −4% (bytes **−22%**) |
| `provider/fragmented_512k` | 12 | 9 | bytes −33% |
| `doc_frame/plain`, `overlay/help`, `overlay/ns_filtered` | — | — | unchanged (paths not touched) |

## Notes

- A first attempt at the provider framing validated UTF-8 per line instead of
  once per received region. It cut allocated bytes but made ordinary chunks
  ~3.5% *slower*, because `parse_entry`'s JSON DOM dominates that path — the
  numbers above are the corrected version, which validates once and slices
  borrowed lines out of it.
- A help-search filter rewrite was also tried and reverted: it removed two
  allocations per binding per frame but moved wall time zero (filtering ~150
  short lines is a few µs of a 199 µs frame), and it added an ASCII/Unicode
  fork that was not worth the complexity.
- `wrap_line_window` is covered by a test asserting it returns exactly the
  slice of the full wrap it stands in for, for every skip/take combination —
  the scroll math depends on that agreement.
- Not done: caching document syntax spans by revision (014, `doc_frame/plain`
  is unchanged), the cached help model and viewport-sliced pickers (011's real
  fix, overlapping finding 068), the concurrent selector sweep (019), and
  delivering version metadata asynchronously (020, needs the connect/startup
  path restructured so it can arrive after `App` exists).
- 012's win is capped at ~12% because a huge line is still fully *styled*
  before wrapping and rows above the viewport are still walked for their
  column arithmetic; clipping before styling and cached breakpoints are the
  rest of that finding.


nklmilojevic:
Lets fix the conflicts and merge in main so CI can run green and I'll review it afterwards.

PR #231, opened 2026-09-05
feat: add optional Popeye integration

vyrti:
## Summary

- add `:popeye` only when standalone or Krew Popeye is detected, and rescan PATH on `:reload`
- stream Popeye JSON directly into a searchable report without buffering the full output
- cap reports at 10,000 lines / 4 MiB, avoid redundant allocations, and kill scans after five minutes

## Testing

- `cargo fmt --all -- --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test` (612 passed, 1 ignored)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2883,7 +2883,6 @@ dependencies = [
  "similar",
  "terminal-colorsaurus",
  "tokio",
- "tokio-util",
  "toml",
  "unicode-width",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2883,6 +2883,7 @@ dependencies = [
  "similar",
  "terminal-colorsaurus",
  "tokio",
+ "tokio-util",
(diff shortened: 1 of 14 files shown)


nklmilojevic:
Thanks for the work on this and the care put into testing it. I’m keeping these integrations outside sofka’s core so they can be maintained and released independently.

The plugin architecture is now merged in #237 and is available in v0.24.0. It provides a shared way to run external tools and display their reports, with input validation, timeouts, cancellation, and guardrails.

I’m closing this PR, but I’d welcome this integration as an external plugin. These resources should help:

- [Plugin authoring guide](https://github.com/nklmilojevic/sofka/blob/main/docs/plugin-authoring.md)
- [Example plugin package](https://github.com/nklmilojevic/sofka/tree/main/examples/plugins/resource-summary)

Thanks again for contributing.

PR #235, opened 2026-09-06
fix: scroll the help panel

elidrissi-amine:
The help panel clips bindings below the visible area. Users cannot browse the full list, including configured plugins, bookmarks, and workspaces.

Help now supports line scrolling with `j`/`k` and arrows, paging with `Ctrl-F`/`Ctrl-B`, `PgDn`/`PgUp`, and `Space`, and top/bottom navigation with `g`/`G` and `Home`/`End`. Each page uses the rendered content height, so small terminals do not skip bindings. The renderer updates the page size and scroll limit after resizing or changing compact mode. Resize events refresh the layout immediately, so a page key before the next frame timer uses the new dimensions. The title shows a scroll hint when content exceeds the visible area.

Opening help, starting a filter, and clearing a filter reset the scroll position. The help text, key reference, and feature list describe the controls.

Validation: tests render help at several terminal sizes and send navigation keys through `handle_key`. They cover paging, scroll limits, filtering, reopening, resizing, and compact mode. `just check` passes with 1,066 tests passed and two ignored.


src/app/input.rs
```diff
@@ -1244,17 +1245,50 @@ impl App {
     }
 
     pub(super) fn key_help(&mut self, key: KeyEvent) {
+        // The help text runs past the bottom of any terminal, so it scrolls
+        // with the same keys as the document views (j/k, arrows, ctrl-f/b,
+        // space, g/G). `help_max_scroll` is set by the renderer, which is the
+        // only place the wrapped line count is known.
+        let page = 10u16;
```
nklmilojevic: Use the help viewport height as the page size. The fixed value of 10 can skip hidden bindings when the viewport has fewer than 10 content rows. For example, an 8-row viewport shows lines 0 through 7. Page Down then starts at line 10, and lines 8 and 9 do not appear. Store the viewport height in `draw_help`. Use that value for Page Down, Space, Ctrl-F, Page Up, and Ctrl-B. Add a test that uses a viewport height of less than 10.

nklmilojevic (changes requested):
Please correct the paging behavior before merge.

PR #239, opened 2026-09-06
feat: ship sanitize as a core plugin

lucsoft:
Rewritten from the ground up for the review. This is no longer an example package — it is a plugin sofka ships, maintains, and installs with itself, with a Rust adapter per the [language policy](https://github.com/nklmilojevic/sofka/blob/main/docs/plugin-authoring.md#language-choice).

`:sanitize` in the pods view deletes the pods a namespace has finished with.

```
:sanitize
:sanitize states=stuck
:sanitize dry_run=true
```

### How it is bundled

The adapter is the sofka binary: `sofka --plugin-adapter sanitize` (hidden from `--help`) speaks the existing request/report protocol on stdin and stdout. `plugins/sanitize/plugin.toml` is embedded with `include_str!` and registered at config load with `command` pointed at the running executable.

So a normal install has nothing to copy and needs no runtime on `PATH` — and it behaves identically for `cargo install`, Nix, and the release binaries, with no `package.nix` or tarball-layout changes. I avoided a `$out/share` data directory precisely because it silently does nothing for `cargo install`.

Registration runs *after* user packages, so an inline `[[plugins]]` entry or a user package named `sanitize` still wins and can replace the shipped one.

Going through the plugin interface means guardrails (`plugin:sanitize`), read-only mode, the ⚠ confirmation, the action journal, output limits, and `:plugin-cancel` all apply with no new machinery.

### Selection reuses the pods view's own STATUS

`columns::pod_status` is now exposed and `pod_summary` delegates to it, so there is one derivation rather than two that drift.

This turned out to matter for correctness, not just tidiness. **The k9s vocabulary does not survive the move**: sofka prints `Succeeded` where k9s prints `Completed`, and `Failed` where k9s prints `Evicted`. The Python version carried the k9s names, so ported literally the two commonest cleanup targets — finished jobs and evicted pods — would never have matched anything.

| `states`   | Deletes |
| ---------- | ------- |
| `terminal` (default) | `Succeeded`, `Failed`, `Error`, `OOMKilled`, `ContainerStatusUnknown` |
| `stuck`    | the above, plus `CrashLoopBackOff`, `ImagePullBackOff`, `ErrImagePull` |
| `all`      | the above, plus `Pending` — the set k9s uses |

`all` is not the default because a `Pending` pod is usually waiting for a node rather than failed.

### The delete race is now actually closed

Earlier in this PR I said an atomic identity-bound delete was impossible; that was true *of kubectl*, which has no UID precondition and whose `delete --raw` takes no body. Through the API it is one field. Every delete carries `Preconditions { uid }`, so a name that now holds a different object is rejected by the server with 409 and counted as replaced rather than deleted. A `409`/`404` is treated as "not ours any more", not as a failure.

A pod is also spared when it is already `Terminating`, or when any container is still running — readiness deliberately not consulted, since a container failing its readiness probe is out of the load balancer but very much alive.

### Tests

`cargo test`, so they run in CI as-is — no Python test infrastructure needed. Unit tests in `src/sanitize.rs` cover the state sets, the status vocabulary (explicitly asserting `Succeeded`/`Failed`, not the k9s names), terminating pods, live-container protection including the unready case, and row capping. Two `handle_key` tests in `src/app/tests.rs` cover the user-visible behaviour: that `:sanitize` is present with no configuration, is scoped to pods, confirms instead of deleting, and is blocked in read-only mode.

Three existing plugin tests asserted exact plugin counts; they now count non-bundled plugins, so they keep testing what they meant to test.

`just check` is green: `cargo fmt --all -- --check`, `cargo test --locked` (687 pass), and clippy clean across every file this branch touches.

### Docs

`plugins/sanitize/README.md` covers requirements, inputs, dry run, deletion states, confirmation, read-only, guardrails, and scope — including, stated plainly, that it ignores the view filter and acts across all namespaces when the view does. Plus a `## Bundled plugins` section in `docs/features.md`, a row in the `docs/keys.md` command table, and a pointer in `docs/plugins.md`. `?` help picks it up automatically, so `src/ui.rs` needed no change.

Rebased onto current main (`fdb58c0`).

### DIFF
--- a/plugins/sanitize/plugin.toml
+++ b/plugins/sanitize/plugin.toml
@@ -1,0 +1,24 @@
+schema_version = 1
+
+[plugin]
+name = "Sanitize pods"
+palette = "sanitize"
+# Rewritten to the running executable when sofka registers this package, so a
+# normal install needs no adapter file and no extra language runtime.
+command = "sofka"
+args = ["--plugin-adapter", "sanitize"]
+scopes = ["pods"]
+target = "context"
+output = "report"
+mutating = true
+dangerous = true
+timeout = "2m"
+
+[plugin.inputs.states]
+type = "string"
+default = "terminal"
+choices = ["terminal", "stuck", "all"]
+
+[plugin.inputs.dry_run]
+type = "boolean"
+default = "false"
--- a/src/sanitize.rs
+++ b/src/sanitize.rs
@@ -91,16 +91,22 @@ pub async fn run() -> Result<()> {
 /// The `[[guardrails]]` that apply to this run. The adapter is sofka, so it
 /// reads the same configuration the session did rather than being told.
 fn guardrails_for(request: &Value) -> Vec<crate::config::Guardrail> {
-    let context = request
-        .get("context")
-        .and_then(Value::as_str)
-        .unwrap_or_default();
+    let context = guardrail_context(request);
     let cluster = request
         .get("cluster")
         .and_then(Value::as_str)
         .unwrap_or_default();
     let (loader, _) = crate::config::ConfigLoader::load();
     loader.resolve(context, cluster).config.guardrails
+}
+
+fn guardrail_context(request: &Value) -> &str {
+    // Cluster::connect uses this name for guardrails without a named context.
+    // Keep the request context null so client_for still uses Config::infer.
+    request
+        .get("context")
+        .and_then(Value::as_str)
+        .unwrap_or("default")
 }
 
 async fn sanitize(
@@ -179,10 +185,7 @@ async fn sanitize(
         .collect();
     let limits = crate::app::guardrails::restrictions(
         guardrails,
-        request
-            .get("context")
-            .and_then(Value::as_str)
-            .unwrap_or_default(),
+        guardrail_context(request),
         "plugin:sanitize",
         "pods",
         &scoped,
@@ -731,6 +734,104 @@ mod tests {
             "{summary}"
         );
         assert_eq!(report["sections"][1]["title"], "Blocked");
+    }
+
+    #[tokio::test]
+    async fn a_context_bulk_limit_applies_without_a_named_context() {
+        for context in [Value::Null, json!("default"), json!("production")] {
+            let guardrail = crate::config::Guardrail {
+                contexts: vec![context.as_str().unwrap_or("default").into()],
+                actions: vec!["plugin:sanitize".into()],
+                max_bulk: Some(1),
+                ..Default::default()
+            };
+            let (client, seen) = mock_api(vec![(
+                "200 OK",
+                list_of(vec![
+                    failed("job-1", "uid-1", "rv-10"),
+                    failed("job-2", "uid-2", "rv-11"),
+                ]),
+            )])
+            .await;
+            let mut request = request("terminal", false);
+            request["context"] = context;
+            let report = sanitize(client, &request, &[guardrail]).await.unwrap();
+            assert_eq!(seen.lock().unwrap().len(), 1);
+            assert!(summary_of(&report).contains("exceeds the guardrail limit of 1"));
+            assert_eq!(report["sections"][1]["title"], "Blocked");
+        }
+    }
+
+    #[tokio::test]
+    async fn a_delete_conflict_or_missing_pod_is_not_counted_as_deleted() {
+        for (status, code, expected) in [
+            ("409 Conflict", 409, "1 changed since the scan"),
+            ("404 Not Found", 404, "1 already gone"),
+        ] {
+            let (client, seen) = mock_api(vec![
+                ("200 OK", list_of(vec![failed("job-1", "uid-1", "rv-10")])),
+                ("200 OK", failed("job-1", "uid-1", "rv-11").to_string()),
+                (
+                    status,
+                    json!({"kind": "Status", "status": "Failure", "code": code}).to_string(),
+                ),
+            ])
+            .await;
+            let report = sanitize(client, &request("terminal", false), &[])
+                .await
+                .unwrap();
+            let calls = seen.lock().unwrap().clone();
+            assert_eq!(calls.len(), 3);
+            assert_eq!(calls[2].0, "DELETE");
+            let summary = summary_of(&report);
+            assert!(summary.contains("0 deleted, 0 failed"), "{summary}");
+            assert!(summary.contains(expected), "{summary}");
+            assert_eq!(report["sections"].as_array().unwrap().len(), 1);
+        }
+    }
+
+    #[tokio::test]
+    async fn a_failed_pre_delete_read_skips_the_pod_and_continues() {
+        let (client, seen) = mock_api(vec![
+            (
+                "200 OK",
+                list_of(vec![
+                    failed("job-1", "uid-1", "rv-10"),
+                    failed("job-2", "uid-2", "rv-11"),
+                ]),
+            ),
+            (
+                "403 Forbidden",
+                json!({"kind": "Status", "status": "Failure", "code": 403,
+                "reason": "Forbidden", "message": "cannot read job-1"})
+                .to_string(),
+            ),
+            ("200 OK", failed("job-2", "uid-2", "rv-11").to_string()),
+            (
+                "200 OK",
+                json!({"kind": "Status", "status": "Success"}).to_string(),
+            ),
+        ])
+        .await;
+        let report = sanitize(client, &request("terminal", false), &[])
+            .await
+            .unwrap();
+        let calls = seen.lock().unwrap().clone();
+        assert_eq!(calls.len(), 4);
+        let deletes: Vec<_> = calls
+            .iter()
+            .filter(|(method, ..)| method == "DELETE")
+            .collect();
+        assert_eq!(deletes.len(), 1);
+        assert_eq!(
+            deletes[0].1.split('?').next(),
+            Some("/api/v1/namespaces/default/pods/job-2")
+        );
+        assert!(summary_of(&report).contains("1 deleted, 1 failed"));
+        assert_eq!(report["sections"][1]["title"], "Failed");
+        assert_eq!(report["sections"][1]["rows"][0][1], "job-1");
+        assert_eq!(report["sections"][2]["title"], "Deleted");
+        assert_eq!(report["sections"][2]["rows"][0][1], "job-2");
     }
 
     #[tokio::test]
--- a/src/sanitize.rs
+++ b/src/sanitize.rs
@@ -8,7 +8,7 @@
 
 use anyhow::{Context, Result};
 use k8s_openapi::api::core::v1::Pod;
-use kube::api::{Api, DeleteParams, ListParams, Preconditions};
+use kube::api::{Api, ApiResource, DeleteParams, ListParams, Preconditions};
 use kube::core::DynamicObject;
 use kube::{Client, Config};
 use serde_json::{Value, json};
@@ -29,6 +29,9 @@ const UNSTARTED: &[&str] = &["Pending"];
 
 /// Rows per report table. The runner caps captured output at 1 MiB.
 const MAX_ROWS: usize = 500;
+/// Pods fetched per list request. Sanitizing all namespaces on a large cluster
+/// is otherwise one unpaged response holding the entire pod inventory.
+const PAGE: u32 = 500;
 
 fn wanted(states: &str) -> Option<Vec<&'static str>> {
     let mut set = TERMINAL.to_vec();
@@ -58,20 +61,11 @@ fn wanted(states: &str) -> Option<Vec<&'static str>> {
 /// counting them would make an identical crash-looping pod exempt from
 /// `states = stuck` purely because something injected a proxy into it. The
 /// STATUS column ignores them for the same reason, so the two agree.
-fn running(pod: &Pod) -> bool {
-    pod.status
-        .as_ref()
-        .and_then(|s| s.container_statuses.as_ref())
-        .is_some_and(|cs| {
-            cs.iter()
-                .any(|c| c.state.as_ref().is_some_and(|s| s.running.is_some()))
-        })
-}
-
-/// A pod's status as the pods view shows it, so `states` names what the user reads.
-fn status_of(pod: &Pod) -> Result<String> {
-    let object: DynamicObject = serde_json::from_value(serde_json::to_value(pod)?)?;
-    Ok(crate::columns::pod_status(&object))
+fn running(pod: &DynamicObject) -> bool {
+    pod.data
+        .pointer("/status/containerStatuses")
+        .and_then(Value::as_array)
+        .is_some_and(|cs| cs.iter().any(|c| c.pointer("/state/running").is_some()))
 }
 
 struct Target {
@@ -89,7 +83,31 @@ pub async fn run() -> Result<()> {
     if request.get("schema_version").and_then(Value::as_u64) != Some(1) {
         anyhow::bail!("unsupported request schema_version");
     }
+    let client = client_for(request.get("context").and_then(Value::as_str)).await?;
+    let report = sanitize(client, &request, &guardrails_for(&request)).await?;
+    serde_json::to_writer(std::io::stdout().lock(), &report).context("writing the report")
+}
 
+/// The `[[guardrails]]` that apply to this run. The adapter is sofka, so it
+/// reads the same configuration the session did rather than being told.
+fn guardrails_for(request: &Value) -> Vec<crate::config::Guardrail> {
+    let context = request
+        .get("context")
+        .and_then(Value::as_str)
+        .unwrap_or_default();
+    let cluster = request
+        .get("cluster")
+        .and_then(Value::as_str)
+        .unwrap_or_default();
+    let (loader, _) = crate::config::ConfigLoader::load();
+    loader.resolve(context, cluster).config.guardrails
+}
+
+async fn sanitize(
+    client: Client,
+    request: &Value,
+    guardrails: &[crate::config::Guardrail],
+) -> Result<Value> {
     let inputs = request.get("inputs").cloned().unwrap_or_else(|| json!({}));
     let states = inputs
         .get("states")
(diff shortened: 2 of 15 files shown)


lucsoft:
Pushed 75e8263 for both review findings. Both were real, and neither was caught by the fixture I had — added cases for both.

**Running pods remain terminal** — correct, and the fix is at the selection, not the status. `pod_status` is a port of k9s's `render.PodStatus`, and reporting the last terminated container's reason is what kubectl's own STATUS column does, so I did not want to diverge there: the report should show the same string the pods view shows. What was wrong was treating that string as sufficient grounds to delete. The scan now skips any pod with a running, ready container, so a pod that reads `OOMKilled` because a sidecar died is left alone while it still serves.

**Deletion loses pod identity** — also correct, and worse than it looks for stable-named pods. Deployment and Job pods get a fresh suffix, but a StatefulSet recreates `db-0` as `db-0`, so `kubectl delete pod db-0` between the scan and the delete takes the replacement. The window is not theoretical at scale: with several thousand pods, the last namespace's delete can be a long way behind the scan that judged it.

kubectl has no UID precondition on `delete`, so this is a re-check rather than an atomic guard: the scan keeps each pod's UID, and immediately before deleting a namespace the adapter re-reads that namespace's `name=uid` pairs and drops any name whose object changed. That takes the window from the length of the whole run down to one round trip. The report counts what the check held back, and a failed re-read deletes nothing in that namespace rather than falling through to a name-only delete.

I would not call it a security issue — a pod is ephemeral and the controller recreates what it owns, so the worst case is one unnecessary restart — but it is a correctness bug and the package is meant to be exemplary, so it is fixed rather than documented.

Both are covered in the fixture now: a multi-container pod with one live and one OOMKilled container, and a StatefulSet pod whose UID changes at the re-check. Full sweep across the three `states` values, dry-run and real, plus the partial-failure and failed-re-read paths, still passes, and `--validate-plugin` / `--validate-plugin-report` are green.

examples/plugins/sanitize/adapter.py
```diff
@@ -0,0 +1,169 @@
...
+    if (pod.get("metadata") or {}).get("deletionTimestamp"):
+        # A pod on a lost node keeps its containers; everything else is on its way out.
+        reason = "Unknown" if status.get("reason") == "NodeLost" else "Terminating"
+    return reason
+
+
+def delete(kubectl, namespace, names):
+    """Delete a batch, retrying one by one so a failure names the pod it belongs to."""
+    argv = ["delete", "pod", "-n", namespace, "--ignore-not-found", "--wait=false"]
+    result = run(kubectl + argv + names)
```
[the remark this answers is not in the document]
lucsoft: Fixed in 1080087 and tightened in e93c67f: the scan now keeps each pod's UID and the namespace's `name=uid` pairs are re-read immediately before each delete, so a name whose object changed since the scan is held back and counted in the report. Still a re-check rather than a precondition — see the non-atomic thread.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

examples/plugins/sanitize/adapter.py
```diff
@@ -0,0 +1,169 @@
...
+                reason = terminated["reason"]
+            elif terminated:
+                signal = terminated.get("signal")
+                code = terminated.get("exitCode", 0)
+                reason = f"Signal:{signal}" if signal else f"ExitCode:{code}"
+            elif container.get("ready") and state.get("running"):
+                running = True
+        if reason == "Completed" and running:
+            ready = any(c.get("type") == "PodReady" and c.get("status") == "True"
+                        for c in conditions)
```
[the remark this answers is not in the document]
lucsoft: Fixed in 1080087, at the selection rather than the status: `pod_status` stays a faithful port of k9s's `render.PodStatus` so the report shows the same string the pods view does, and the scan now skips any pod with a running, ready container. A pod reading `OOMKilled` because a sidecar died is left alone while it still serves.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

examples/plugins/sanitize/adapter.py
```diff
@@ -0,0 +1,200 @@
...
+        # A pod name is reusable, so a StatefulSet can put a healthy pod behind the
+        # name that was dead during the scan. Delete only the objects that scanned.
+        current = uids(kubectl, ns)
+        if current is None:
+            failed.append([ns, "", "could not re-read pod UIDs; nothing deleted"])
+            continue
+        names = sorted(name for name, _, uid in targets[ns] if current.get(name) == uid)
+        replaced += len(targets[ns]) - len(names)
+        for start in range(0, len(names), BATCH):
+            done, bad = delete(kubectl, ns, names[start:start + BATCH])
```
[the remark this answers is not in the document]
lucsoft: Correct, and I have narrowed it rather than closed it — leaving this thread open because closing it is your call, not mine.

e93c67f moves the identity check from once per namespace to immediately before each delete command, including the individual retries. That matters more than it sounds: with 120 dead pods the old code checked once and then deleted three batches against that one reading, so the last batch was deleted long after it was cleared. A test where a pod is replaced after the first batch now holds it back (119 deleted, 1 held) where the previous code deleted it.

What is left is genuinely not closable with kubectl. `kubectl delete` has no UID or resourceVersion precondition, and `delete --raw` takes no body, so DeleteOptions preconditions cannot be sent. The delete is name-addressed no matter how carefully the target was chosen. `--field-selector` does not help either: kubectl resolves it by listing then deleting by name client-side, and a replacement pod matches `status.phase=Pending` anyway.

Closing it properly means talking to the API directly with `preconditions.uid`, which means a real Kubernetes client instead of kubectl. That is a fair thing to want in a destructive plugin, but it changes this from a dependency-light example into something else, so I would rather you decide. Happy to rewrite it that way if you prefer — otherwise the residual window is one round trip and the consequence is a redundant restart of a pod its controller owns.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

examples/plugins/sanitize/adapter.py
```diff
@@ -0,0 +1,212 @@
...
+def live(pod):
+    """Whether a regular container is still serving.
+
+    kubectl and k9s report the last terminated container's reason even when a
+    sibling is running, so a multi-container pod can read `OOMKilled` while it
+    still takes traffic. That is fine for a STATUS column and wrong for a
+    delete, so the status stays faithful and the selection checks this instead.
+    """
+    return any(c.get("ready") and "running" in (c.get("state") or {})
+               for c in (pod.get("status") or {}).get("containerStatuses") or [])
```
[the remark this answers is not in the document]
lucsoft: Taken as suggested in b3d2660. The `ready` conjunct was copied from the shape of the k9s status logic, but that logic is deciding what to print and this is deciding whether to kill a process — a container running but failing its readiness probe, or still inside its startup probe, is out of the load balancer and entirely alive. Readiness is now irrelevant to the guard, and the docstring says why the two checks deliberately differ so nobody harmonises them later. Nothing legitimate is lost: a finished or wedged pod has its containers terminated or waiting, never running. Fixture gained a running-but-unready container alongside a terminated sibling.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nklmilojevic (changes requested):
Please make `sanitize` an official plugin bundled with sofka. Sofka should maintain and distribute this feature through the existing plugin interface.

Before merge, please make these changes:

- Move the package from `examples/plugins/sanitize/` to `plugins/sanitize/`. Keep `examples/plugins/` for examples that teach plugin authors.
- Add the packaging and discovery support needed to make `:sanitize` available with a normal sofka installation, without a manual package copy. Moving the directory alone is not sufficient. Show clear instructions if Python or kubectl is absent.
- Add user documentation and a package README. Cover requirements, inputs, dry-run use, deletion states, confirmation, read-only restrictions, and namespace scope. State clearly that the command ignores the view filter and can act across all namespaces. Update the relevant feature and command documentation.
- Commit the adapter tests and fixtures described in the PR, and run them in CI. Cover live-container protection, pod replacement, partial deletion failures, failed identity checks, and dry-run behavior. Add application tests through `handle_key` for any new user-visible behavior in sofka.
- Make sure `just check` passes before this PR is ready.

The existing plugin interface is a suitable implementation for this feature. A separate Rust implementation is not required. The requested change is to make this a maintained, distributed sofka feature with tests and documentation. An example package alone does not provide that.

nklmilojevic:
The plugin language policy is now on main in https://github.com/nklmilojevic/sofka/commit/fdb58c0.

For core plugins maintained and shipped by sofka, we prefer Rust. This keeps the implementation consistent with sofka and reduces the need for extra language runtimes. Rust is a preference, not a requirement of the plugin protocol.

External plugins can use any programming language that follows the plugin protocol. The Python example demonstrates the protocol; it does not set the language preference for core plugins.

For the sanitize plugin in this PR, Rust is the preferred choice if sofka will maintain and ship it as a core plugin. A separate external package can use Python.

See [Language choice](https://github.com/nklmilojevic/sofka/blob/main/docs/plugin-authoring.md#language-choice).

lucsoft:
Happy to take it in that direction — bundled and maintained rather than an example.

One thing to confirm first, because it decides most of the work.

`docs: prefer Rust for core plugins` (fdb58c0) landed about ten minutes after this review, and I read the two together as: keep the plugin interface, but write the adapter for a core plugin in Rust — with "reduces the need for extra language runtimes" pointed squarely at my `requires = ["python3", "kubectl"]`. I read "a separate Rust implementation is not required" in the review as "do not build this as a built-in that bypasses the plugin system", rather than "do not write the adapter in Rust". Since the policy also says Rust is a preference and not a requirement, please confirm — it is the difference between porting ~200 lines and keeping them.

Assuming that reading, the plan:

**Adapter.** Rust, speaking the existing manifest/request/report protocol over stdin/stdout, invoked as a hidden `sofka` subcommand rather than a second installed binary. That also answers your second bullet: with no data files and no runtime to install, `cargo install`, Nix and the release binaries all behave identically, and the manifest can be embedded and registered at startup ahead of `<config>/plugins/`, still overridable by a user package of the same name. If you would rather have a separate workspace binary, or a real package directory installed into a data dir, say so — I did not pick the data-dir route because it needs `package.nix` and the release tarball layout to change and quietly does nothing for `cargo install`.

**Dependencies.** Both go. kube-rs is already a dependency, so the adapter talks to the API directly, `requires` becomes empty, and there is no "install Python" instruction left to show.

**A real fix that comes with it.** The review bot flagged, correctly, that a name-addressed delete is racy: between judging a pod dead and deleting it, a StatefulSet can put a healthy `db-0` behind the same name. With kubectl that is genuinely unfixable — no UID precondition, and `delete --raw` takes no body — so the current adapter narrows the window to a single round trip and documents the remainder. `DeleteParams` with `Preconditions { uid }` closes it outright. That is on its own a decent argument for the Rust adapter.

**The rest of your list** stands: move to `plugins/sanitize/`; package README plus user documentation covering requirements, inputs, dry run, deletion states, confirmation, read-only, and namespace scope, stating plainly that the command ignores the view filter and can act across all namespaces; the feature and command docs; the adapter tests and fixtures committed and run in CI — which become ordinary `cargo test` cases once the adapter is Rust, removing the "no Python test infrastructure" problem — plus `handle_key` tests for anything user-visible, and `just check` green.

One correction while I am here: the description currently says clippy reports pre-existing findings. That was my local toolchain (1.95.0) running ahead of CI's stable — the Clippy step passes on CI and there is nothing wrong on your side. I will drop that line.

I will rebase onto current main before pushing any of this.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/sanitize.rs
```diff
@@ -0,0 +1,359 @@
...
+/// part of this: a container failing its readiness probe, or still inside its
+/// startup probe, is out of the load balancer and very much alive.
+fn running(pod: &Pod) -> bool {
+    pod.status
+        .as_ref()
+        .and_then(|s| s.container_statuses.as_ref())
+        .is_some_and(|cs| {
+            cs.iter()
+                .any(|c| c.state.as_ref().is_some_and(|s| s.running.is_some()))
+        })
```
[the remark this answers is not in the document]
lucsoft: Real gap, but I did not take the suggested change — counting init containers would make the feature incoherent, so I pinned the current behaviour instead (65de5cb).

Native sidecars land in `initContainerStatuses` and run for the pod's lifetime, so including them means a crash-looping pod is spared when a mesh injected a proxy and deleted when it did not — identical workloads, opposite outcomes, decided by something unrelated to whether the pod is alive. `states=stuck` would effectively stop working on any meshed cluster. `columns::pod_status` ignores init containers for the same reason, so selection and the STATUS column stay in agreement.

The benign case turns out to need no rule: once the application container finishes cleanly the status reads `Running` (the terminated reason `Completed` is skipped and the phase shows through), which is in no state set. So the only situation your scenario reaches is an application container that is dead or wedged while its proxy is still up, and there the proxy is infrastructure for a workload that is going nowhere.

What was genuinely wrong was the wording: `running()` said "any container" and the README said "a pod with a running container", both of which promised more than the code does. Now stated as "a running application container", with the reasoning in the function doc and a test covering both halves — sidecar plus crash-loop is selected, sidecar plus completed is not.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/sanitize.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    if !dry_run {
+        for target in &targets {
+            // The UID precondition binds the delete to the object that scanned.
+            // A pod name is reusable, so without it a StatefulSet that recreated
+            // `db-0` in the meantime would lose its healthy replacement instead.
+            let params = DeleteParams {
+                preconditions: Some(Preconditions {
+                    uid: Some(target.uid.clone()),
+                    resource_version: None,
```
nklmilojevic: [P1] Check the scanned resource version before deletion

The UID protects a replacement pod, but it does not protect the same pod if its status changes. For example, `states=stuck` can select a pod in `CrashLoopBackOff`, then its application container can restart and become ready before this loop reaches it. The UID remains the same, so this delete can remove the recovered pod.

I reproduced this with the PR binary and a local mock API: LIST returned resource version 10 with `CrashLoopBackOff`; before DELETE, the pod changed to version 11 with a running, ready container. The adapter sent only the UID precondition and reported the pod as deleted.

Please retain each scanned pod's `resourceVersion` and include it with the UID in the delete preconditions. Skip pods whose preconditions fail, and report a changed pod accurately. Add a regression test for recovery between LIST and DELETE.

plugins/sanitize/plugin.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[plugin]
+name = "Sanitize pods"
+palette = "sanitize"
+# Rewritten to the running executable when sofka registers this package, so a
+# normal install needs no adapter file and no extra language runtime.
+command = "sofka"
+args = ["--plugin-adapter", "sanitize"]
+scopes = ["pods"]
+target = "context"
```
nklmilojevic, line 11: [P1] Enforce the bulk limit against the pods to delete

With `target = "context"`, `run_plugin()` creates one target for `guard_scope()`. The guardrail compares `max_bulk` with that one target, while the adapter independently lists and deletes every matching pod. Thus, a rule with `actions = ["plugin:sanitize"]` and `max_bulk = 1` still permits deletion of many pods. This conflicts with the README's statement that guardrails can cap this operation.

Please enforce the applicable limit against the actual deletion set before the first DELETE. Add a test through `handle_key` for the guardrail path and an adapter test that proves an over-limit set causes no deletions.

src/sanitize.rs
```diff
@@ -0,0 +1,406 @@
...
+                .await
+                .with_context(|| format!("building config for context '{name}'"))?
+        }
+        None => Config::infer().await.context("loading kubeconfig")?,
+    };
+    Client::try_from(config).context("building a Kubernetes client")
+}
+
+#[cfg(test)]
+mod tests {
```
nklmilojevic, line 378: [P2] Add tests for adapter execution and failure paths

These tests check status helpers and report row limits, but none executes the list/delete path. The application tests stop before adapter execution. As a result, the tests do not verify the outgoing UID precondition, handling of 404/409 responses, partial deletion failures, failed LIST requests, or that `dry_run=true` sends no DELETE requests. These cases were requested in the earlier review.

Please add tests with a mock Kubernetes API that execute the adapter path and check both requests and reports. Include the recovery race described above, and keep these tests in the normal `cargo test --locked` suite.

nklmilojevic (changes requested):
The bundled Rust plugin addresses the distribution requirement. However, changes are still required before merge: protect pods that recover after the scan, enforce the documented bulk limit against the actual pod count, and add the requested adapter execution tests. The inline comments give the details.

Validation on commit 65de5cb: `just check` passed, with 688 tests passed and one ignored. A local mock API test confirmed that the adapter sends a UID-only delete after the same pod becomes running and ready. GitHub CI is still in `action_required` state; it must pass before merge.

src/sanitize.rs
lucsoft: Race confirmed and fixed in ed244ca, but not with the scanned `resourceVersion` — I think that instrument would fix the race and break the command, so I took the semantic route instead. Happy to switch if you disagree.

The problem with pinning the version from the list: the statuses that can recover are exactly the ones whose `resourceVersion` churns hardest. `Succeeded` and `Failed` are absorbing phases and mostly quiet, but `CrashLoopBackOff`, `ImagePullBackOff`, `Error` and `OOMKilled` all bump on every restart, backoff tick and probe update — for reasons that have nothing to do with recovery. A run over a few hundred crash-looping pods would refuse a large share of them as "changed", none of which had been anything but dead, and the user would just run it again. That trades a rare wrong delete for a routinely useless command.

So instead each pod is re-read immediately before its delete and the selection predicate is re-run against what comes back: same UID, status still in the set, still nothing running. That asks the question we actually mean — did this stop qualifying? — and is blind to churn that does not change the answer. The delete then carries preconditions for the UID **and the version from that read**, so the precondition is tight rather than stale, and the remaining window is one round trip instead of the whole run.

Cost is one GET per pod, which I have noted in the README as a reason a very large namespace needs a longer timeout.

Regression tests are in `src/sanitize.rs` against a mock apiserver: a pod that recovers between list and delete never reaches DELETE; a replacement under the same name never reaches DELETE; and a pod whose restart count and version moved while it stayed dead is still deleted — that last one is the case the scanned-version approach would have skipped.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

plugins/sanitize/plugin.toml
lucsoft: Fixed in ed244ca — you were right that the README was writing a cheque the code did not honour.

`crate::app::guardrails::restrictions()` is now a free function over `&[Guardrail]` rather than living on `App`, so `guard_scope` and the adapter share one implementation instead of growing a second copy that drifts. The adapter re-evaluates `plugin:sanitize` against the set it actually matched, before the first DELETE, and when the set is over the limit it deletes nothing at all and says so in the report with a `Blocked` table rather than partially applying.

Tests: an adapter test with `max_bulk = 1` and two matching pods asserts exactly one request is made — the list — and no DELETE; a companion test with `max_bulk = 2` and one pod confirms the limit does not block a set within it. Plus a `handle_key` test for the app-side path, where a `deny` rule on `plugin:sanitize` stops it before the confirmation dialog.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/sanitize.rs
lucsoft: Added in ed244ca. There is now a mock apiserver in `src/sanitize.rs` tests that answers a scripted sequence and records the method, path and body of every request, so the assertions are about what the adapter actually sent, not just what it reported. Ordinary `cargo test --locked`, no new dev-dependency — it follows the TCP-listener pattern already used by the watch tests in `src/k8s.rs`.

Covered: the outgoing preconditions on DELETE (UID plus the version from the pre-delete read); recovery between list and delete; a replacement wearing the same name; churn that moves the version without changing the verdict, which must still delete; 404 counted as already gone rather than deleted; a 403 on one pod not hiding the success of another; a failed LIST surfacing as an execution error instead of an empty report; `dry_run=true` issuing exactly one GET and no DELETE; paging following the continue token; and the bulk limit blocking before any DELETE.

Two things came out of writing them, both now fixed. The scan was `Api::all` with default `ListParams` — one unpaged response holding every pod in the cluster — and it round-tripped each pod through serde to get a status. It now pages with a continue token, keeps only the few strings per target, and lists `DynamicObject` so `columns::pod_status` reads it directly.

Separately, the view filter is no longer ignored. `-l` and `-f` terms are Kubernetes selectors and now narrow the scan server-side; the rest of the grammar is evaluated against rendered cells inside the app and cannot be reproduced here, so rather than sanitize a wider set than the table shows, a filter like `web-` refuses to run and says why. That was a footgun I had only documented.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nklmilojevic:
I pushed 3f369c5 to address the remaining review findings.

Without a named kubeconfig context, the app uses `default` for guardrails, but the adapter used an empty string. This caused the adapter to miss context-specific bulk limits. I reproduced this with a mock API: `contexts = ["default"]` and `max_bulk = 1` still allowed two pod deletions.

The adapter now uses the same guardrail context as the app for configuration loading and bulk-limit checks. The Kubernetes client still uses normal configuration inference when the request has no named context.

I also added regression tests for context-specific limits, DELETE responses 404 and 409, and a failed pre-delete read. The read-failure test checks that the affected pod is not deleted and that processing continues for the next pod. An application test checks the guardrail path through `handle_key`.

`just check` and the commit hooks passed: 710 tests passed and one was ignored.

lucsoft:
Thanks — that was my bug and a good catch. `Cluster::connect` falls back to `default` for the guardrail context and I passed the raw request value straight through, so a `contexts = ["default"]` rule matched in the app and silently missed in the adapter. Worse than a plain miss, since the whole point of that code path is enforcing a limit the runner cannot. Splitting `guardrail_context()` from what `client_for` receives is the right shape: the guardrail name and the client's context inference genuinely are two different questions, and I had conflated them.

I have pulled 3f369c5 down, fast-forwarded, and verified locally: `cargo fmt --all -- --check` clean, 710 passed and one ignored, no clippy findings in the touched files. Your test looping the same guardrail over `null`, `"default"` and `"production"` is a better shape than the single case I would have written.

I have force-pushed this branch twice earlier in the review; now that it carries your commits I will not do that again — anything further from me goes on top.

One thread is still open: the `resourceVersion` one, where I did not do what was asked and argued for a pre-delete re-read instead. Your new test for a failed pre-delete read suggests you have taken that design, but I would rather you close the thread than assume it. If you would still prefer the scanned version pinned as a precondition, say so and I will switch it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #313, opened 2026-09-07
fix: show SchedulingDisabled in node status when unschedulable should show as yellow

steffen-karlsson:
The node STATUS column now shows `Ready, SchedulingDisabled` as yellow (healthy but not accepting pods)

Screenshot 2026-09-07 at 10 42 53

### DIFF
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -2395,7 +2395,7 @@ async fn cordoned_node_statuses_keep_readiness_colors() {
     );
     assert_eq!(
         cell_color(&terminal, status_start, status_end, ready_y),
-        theme::green()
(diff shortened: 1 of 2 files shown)


nklmilojevic:
hm @steffen-karlsson I think I just fixed this one and released a new version, can you please check it out? Should be obsolete :)

nklmilojevic:
Found it! #259 already added SchedulingDisabled to the node status and is now on main. The yellow status for a cordoned Ready node is a good addition, though. Could you update this PR against the latest main and keep just that color change, with a test? NotReady should stay red.

steffen-karlsson:
Done @nklmilojevic  🥳

PR #319, opened 2026-09-07
feat: name the cluster in the startup banner

purushottam-tuladhar-covergo:
## What

The pre-connect banner said only `Connecting to cluster…`. It now names the cluster it is about to contact:

```
$ sofka
Connecting to cluster: prod-eu…
```

## Why

With several kubeconfig contexts in play, a stale current-context is easy to miss, and the old banner gave no hint about which cluster was being contacted.

The banner is printed before connecting (deliberately — errors stay readable before the TUI takes over the terminal), so the name has to come from the kubeconfig rather than from the live connection. The existing offline helpers `current_context_info` and `cluster_name_for_context` already resolve both the current context and the cluster a `--context` argument points at.

## Which name

The **context** name wins over the kubeconfig cluster name: it is what `kubectl config current-context` prints and what people recognise, whereas the cluster name is a full ARN on EKS (`arn:aws:eks:eu-west-1:1234:cluster/prod`) and `gke_project_zone_name` on GKE.

Three cases:

| kubeconfig state | banner |
| --- | --- |
| context known (current, or `--context foo`) | `Connecting to cluster: foo…` |
| context unknown, cluster name known | `Connecting to cluster: <cluster>…` |
| neither (in-cluster config, no kubeconfig) | `Connecting to cluster…` (unchanged) |

## Tests

`connect_label` is a pure function over `(context, cluster_name)` with unit tests for all three cases in `src/k8s.rs`, keeping the kubeconfig I/O in `main`. `cargo fmt --check` and `cargo test` are clean (894 tests).

Note: `cargo clippy -D warnings` reports 10 `collapsible_match`/boolean-simplification warnings on `main` with the flake's current Rust 1.95 toolchain, all in `src/app/*` and none in the files this PR touches — pre-existing, not from this change.

### DIFF
--- a/src/k8s.rs
+++ b/src/k8s.rs
@@ -537,6 +537,19 @@ pub fn current_context_info() -> Option<(String, String, String)> {
     Some((context, cluster_name, server))
 }
 
+/// Name for the pre-connect startup banner. The kubeconfig context is what
+/// `kubectl config current-context` prints and what users recognise, while the
+/// cluster name is an ARN on EKS and `gke_project_zone_name` on GKE, so the
+/// context wins and the cluster name is the fallback. `None` when the
+/// kubeconfig knows neither (in-cluster config, no kubeconfig).
+pub fn connect_label(context: &str, cluster_name: &str) -> Option<String> {
+    match (context.trim(), cluster_name.trim()) {
+        ("", "") => None,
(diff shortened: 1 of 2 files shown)


nklmilojevic:
This is not needed as you see the name of the cluster as soon as you connect.

PR #320, opened 2026-09-07
feat: drill from MachineDeployment into its Machines

steffen-karlsson:
Pressing enter on a MachineDeployment now lists the Machines it manages, using the same `spec.selector.matchLabels` mechanism as Deployment → Pods. Previously enter fell through to the YAML view, making it hard to navigate the Cluster API hierarchy.

`y` continues to open the YAML view as before.

### DIFF
--- a/docs/keys.md
+++ b/docs/keys.md
@@ -13,7 +13,7 @@ pickers keep both characters available as input.
 | `:<resource> <ns>`                            | switch kind and namespace at once (`:deploy social`; `all`/`*` = all namespaces; the namespace tab-completes)                                                 |
 | `[` / `]`                                     | view history - back / forward through visited kind+namespace views                                                                                            |
 | `Tab` / `shift-Tab`                           | next / previous common resource in the current namespace; cycle workspace views when one is open                                                              |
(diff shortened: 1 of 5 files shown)


nklmilojevic (changes requested):
Thank you, Steffen, for this useful addition and for the tests. This fits sofka's direction well: it makes the Cluster API hierarchy easier to navigate and uses the existing selector and watch code. The qualified Machine target is a good correction too.

There is one remaining API group issue to fix before merge. The new navigation branch matches only `machinedeployments`, so a resource with that plural in another API group also gets the Cluster API action. It could open unrelated Cluster API Machines, or show a warning instead of its normal view. The new header hint and `BUILTIN_DRILLS` entry also apply to every group with that plural; the latter prevents those resources from using a configured drill.

Please restrict this new behavior to `cluster.x-k8s.io` in navigation, hints, and drill configuration validation. Please also add a test through `handle_key` with a `MachineDeployment` in another group. It should retain its configured drill, or open YAML when no drill is configured.

The existing code has some similar plural-only matches, so I understand why you followed that pattern. This request is limited to the new Cluster API behavior. A larger relationship-system change can be separate.

Thanks again for the contribution. This is a useful improvement, and I would like to include it once that group check is in place.

PR #338, opened 2026-09-07
feat: let a drill read the row by JSON Pointer and filter client-side

caseycs:
## What problem this solves

`[views."…"].drill` (#189) covers relationships a label or an indexed field can express, with `{name}` and `{namespace}` from the row's metadata. Two common shapes fall outside that:

- **The target's name sits in the row's `spec`.** A PersistentVolumeClaim names its VolumeAttributesClass in `spec.volumeAttributesClassName`; nothing in metadata carries it, so no template could reach it.
- **Nothing server-side expresses the relationship.** Going the other way, a VolumeAttributesClass's claims carry no label naming it, and the apiserver indexes no PVC field but `metadata.name`/`metadata.namespace`, so a `fields` selector on `spec.volumeAttributesClassName` is rejected by the watch.

## What this does

Two small extensions to `drill`, both generic:

**`{/json/pointer}` placeholders.** Any `{/…}` token in `labels`, `fields`, or `filter` is a JSON Pointer into the row, resolved through the same `extract` that custom columns use. It must land on a non-empty string; a row where it doesn't warns (`/spec/volumeAttributesClassName is empty on scratch`, or `… is not a string`) and stays put, instead of opening a list scoped to nothing.

**`filter`.** A client-side row filter applied after landing, in the `/` key's syntax, filled from the same placeholders — for a target no selector can express. It goes through the same path a bookmark's `filter` does, so `-l`/`-f` terms still reach the server and `esc` behaves as for any filter: first press clears it, second pops back.

```toml
[views.persistentvolumeclaims]
drill = { kind = "volumeattributesclasses", fields = "metadata.name={/spec/volumeAttributesClassName}" }

[views.volumeattributesclasses]
drill = { kind = "persistentvolumeclaims", filter = "{name}" }
```

Nothing about PVCs or VACs is built in; they are the documented example.

## Shape of the change

- `Drill` gains `filter`; `labels_for`/`fields_for` become one `resolve(&obj) -> Result<DrillTarget, String>`, so a bad pointer is one message rather than three silent empties.
- `fill_placeholders` walks `{…}` tokens: `name`, `namespace`, or a pointer via `extract`. `unknown_placeholders` accepts `/`-prefixed tokens, so `compile` still warns on typos like `{node}`.
- `drill_to` returns whether it landed, and `drill_configured` applies the filter only then.
- The drill scope label takes its singular from the kind name rather than `trim_s`, which left `-es` plurals one letter off (`volumeattributesclasse/gold`).

## Tests and docs

- `src/app/tests.rs`, through `handle_key`: PVC → its class by pointer (cluster-scoped target drops the namespace), a PVC without a class warns and stays, VAC → its claims by filter across all namespaces with the two-step `esc`.
- `src/views.rs`: pointer placeholders in `fields` and `filter`, empty and non-string pointers as errors, `filter` compiles, the unknown-placeholder warning names the pointer form.
- `docs/views.md`: the two new stanzas and a paragraph on `filter` and pointer placeholders.

`cargo fmt --all -- --check`, `cargo clippy --locked --all-targets -- -D warnings`, and `cargo test --locked` (895 passing) are clean.

### DIFF
--- a/docs/views.md
+++ b/docs/views.md
@@ -118,8 +118,10 @@ list; a pointer that lands on something other than a string warns that the
 pointer is wrong.
 
 **A row selects other objects** - `drill` names the kind `enter` should open and
-how to scope it: a label selector (`labels`), a field selector (`fields`), or
-both, with `{name}` and `{namespace}` filled in from the row:
+how to scope it: a label selector (`labels`), a field selector (`fields`), a
+client-side row filter (`filter`, the `/` key's syntax), or any mix. Templates
+take `{name}` and `{namespace}` from the row's metadata and `{/json/pointer}`
+from anywhere in the object:
 
 ```toml
 # children carry the parent's name in a label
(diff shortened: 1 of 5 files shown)


caseycs:
Closing in favour of a broader design: an *adjacent* view that lists the objects directly connected to the selected row (owners, children, references, usages) and navigates to them, rather than stretching `drill` to cover single-object references. PR to follow.

PR #342, opened 2026-09-07
feat: extend GitOps view and columns to Argo CD

pujan14:
The GitOps view only understood Flux: an object Argo CD applied reported "not managed by Flux", and Applications listed as bare NAME/AGE rows even though `t` could already suspend, resume, and sync them. Half the feature was there — the actions — with none of the navigation or reading.

`gitops.rs` now resolves an owner for either controller. Argo's stamps are the `argocd.argoproj.io/tracking-id` annotation and the instance labels, which carry the Application's instance name (`<ns>_<name>` outside the controller's namespace, a bare name inside it, so a stamp without a namespace is searched for by name across all of them). The generic `app.kubernetes.io/instance` label is treated as a weak signal: Helm sets it too, so an Application it names that isn't in the cluster reads as unmanaged rather than as a broken chain, and the edit warning stays quiet for it.

From the Application the chain runs to its sources, sync and health status, the revision synced, its sync policy, and the AppProject and generating ApplicationSet beside it — with the same "what is blocking" summary the Flux chain ends on, spelling out drift, degraded health, a failed sync, or a manual policy that will reconcile nothing.

Applications, ApplicationSets, and AppProjects get curated columns. Their plurals are generic, so `columns_for` takes the API group and only claims them inside `argoproj.io`; another `applications` CRD keeps the printer-column fallback. Health colors the row, which needed Argo's status vocabulary (Healthy, Synced, OutOfSync, Missing, Suspended) in the theme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ architecture. The short version:
   patches. No `flux` binary. Plus a native Helm inspector that decodes release
   Secrets itself.
 - **Argo CD built in** - `t` suspends, resumes, and syncs ArgoCD Applications
-  and ApplicationSets through native API patches. No `argocd` binary.
+  and ApplicationSets through native API patches, `:argo` walks the
+  Application → source → revision chain behind any object it applied, and
+  Applications get sync/health columns. No `argocd` binary.
 - **It tells you why something is broken** - `X` opens a deterministic,
   evidence-based incident view. No AI, no external service.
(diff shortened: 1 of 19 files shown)


nklmilojevic:
Please use issues before opening PR - somethings need to be discussed before doing the implementation.

Issue #368, opened 2026-09-08
Discuss new features before opening an issue or pull request, bugs can still be opened as issues


nklmilojevic:
## New features start in Discussions

Please propose new features in GitHub Discussions before you open a feature issue or pull request. Wait for a maintainer to agree on the scope before you start implementation.

Describe the problem, how you handle it now, and how the proposed feature would help. You do not need code or a full design to start a discussion.

Once we agree to proceed, link the discussion in the feature issue and pull request. A feature issue or pull request does not replace this discussion.

## Bug fixes can go directly to a pull request

You can report a bug in an issue or open a bug fix pull request without a discussion first. Include steps to reproduce the problem, what you expected, and what happened.

## If you use a coding agent

Read AGENTS.md before work starts. CLAUDE.md also points to these rules. Agents must check for a feature discussion with maintainer agreement before implementation and show a warning if it is missing.

Review and understand generated code before you submit it. Run the required checks. You are responsible for the change you submit.

These instructions help agents follow the process, but they do not enforce it. Contributors must follow the same process with or without an agent.

PR #381, opened 2026-09-08
feat: read contexts from several kubeconfigs and switch between them

TheiLLeniumStudios:
sofka reads one kubeconfig — the one `kubectl` resolves at startup. This lets the context
switcher read from several at once, so a directory of per-cluster files doesn't mean
restarting with a different `--kubeconfig`. One live cluster at a time; nothing changes for a
single `~/.kube/config`.

A source is a path, and a path is a file or a directory of them. Extra sources come from
`[kubeconfigs] paths`, from `:kubeconfig`, or from `--kubeconfig`, which now repeats and
accepts directories. A directory contributes every kubeconfig under it, re-read whenever the
switcher opens.

Each row shows the context, its server, the namespace it pins, and the kubeconfig it came from
where that isn't already obvious:

context switcher

`:kubeconfig` manages the sources — `a` add, `d` remove. Edits persist to
`<state-dir>/kubeconfigs.toml` and overlay the config rather than rewriting it, like fleet marks:

kubeconfig sources

The header gains a quick-switch strip of clusters visited this session, hidden until there are
two and off with `cluster_strip = false`. Slots are bound to the shifted digits, since macOS
terminals mostly don't transmit `ctrl-<digit>`:

cluster strip

Identity is `(source, context)`, so two kubeconfigs that both define `prod` no longer share one
remembered namespace and one fleet mark. The state key stays the bare name for the default
kubeconfig, so existing state keeps resolving, and a name is qualified as `name@file` only
where another kubeconfig defines it too.

`src/kubeconfigs.rs` is new and pure. Docs and keys are updated; fmt, clippy and tests are
green on v0.24.9.


nklmilojevic:
Please read the guide before opening PRs https://github.com/nklmilojevic/sofka/issues/368

TheiLLeniumStudios:
My bad. Started a new discussion around this

PR #413, opened 2026-09-09
fix: recompute time-dependent sort keys on every rebuild

malpou:
## What

Sorting pods by AGE (ascending, newest first) could land a brand-new pod in the middle of the table instead of at the top. The same applies to running-job DURATION and user `time` columns.

## Why

AGE sort keys are `now - creationTimestamp`, but the sort-key cache validates by `(header, resourceVersion)` and `now` is not part of the key. When a new pod arrived and triggered a rebuild at a later `now`, unchanged pods reused stale keys computed with an earlier `now`. Their cached ages were artificially small, so the new pod's fresh (larger) age sorted after them.

This was especially visible in all-namespaces mode: more pods means more stale cached keys, increasing the chance a new pod's fresh age falls between stale values.

## Fix

- `ViewSpec::is_time_sort` identifies sort columns whose keys depend on `now`: AGE, running-job DURATION, user `time` columns.
- `volatile_sort` extended to include these columns: their sort keys bypass the cache and are always freshly computed (same path as CPU/MEM metrics).
- `time_sensitive` extended so the cache rebuilds once per second even without a watch event.

## Proof of the bug

Two tests that fail on `main` and pass with the fix:

1. `age_sort_new_pod_not_lost_among_stale_cached_keys` - injects stale sort keys (age=10s) for old pods, then applies a new pod (age=30s). Before the fix: `[old-a, old-b, new-pod]` (new pod last). After: `[new-pod, old-b, old-a]`.

2. `age_sort_rebuilds_periodically_without_watch_events` - simulates time passing without a watch event. Before the fix: the cache does not rebuild (time_sensitive was false for sort-only). After: it rebuilds once per second.

### DIFF
--- a/src/app/rows.rs
+++ b/src/app/rows.rs
@@ -467,23 +467,29 @@ impl App {
             return;
         }
         let parsed = self.parsed_filter();
+        let headers = self.display_headers();
+        let sort_header = self
+            .sort_column
+            .and_then(|i| headers.get(i).map(String::as_str));
+        // AGE and other time-dependent sort keys move without a new
+        // resourceVersion, so they can never be cached and the cache must
+        // rebuild once per second even without a watch event.
+        let time_dependent_sort =
+            sort_header.is_some_and(|h| self.spec.is_time_sort(h, &self.kind_plural));
         cache.time_sensitive = match &*parsed {
             crate::filter::ParsedFilter::Structured(s) => {
                 s.terms.iter().any(crate::filter::Term::time_sensitive)
             }
             _ => false,
-        };
+        } || time_dependent_sort;
         cache.filter_second = now;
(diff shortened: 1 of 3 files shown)


nklmilojevic:
Heya! Thanks for this contribution, good catch. I will merge it now and it will be in the next release. Could you in the future just please open up an issue with bug description before you open up a PR? It is all good, just prefer to have an issue before we start working on a bug. Thanks! :)

malpou:
> Heya! Thanks for this contribution, good catch. I will merge it now and it will be in the next release. Could you in the future just please open up an issue with bug description before you open up a PR? It is all good, just prefer to have an issue before we start working on a bug. Thanks! :)

I'll do that next time 👍🏼

PR #414, opened 2026-09-09
ci: add SBOM generation and upload to release workflow

chrxmvtik:
Added a new job to generate and upload a Software Bill of Materials (SBOM) as part of the release workflow.


nklmilojevic:
Please start a discussion as required by #368. Describe the need for an SBOM and how it would be used. We can return to this change after we agree on the scope. I am closing this PR for now.

PR #417, opened 2026-09-09
feat: make mouse wheel scroll distance configurable

### LINKED ISSUE #424: feat: make mouse wheel scroll steps configurable
Mouse wheel events currently cause three navigation steps. This can make selection difficult when a terminal sends several events for a small trackpad movement.

Add a top-level `mouse_scroll_lines` setting to control the number of navigation steps per received wheel event. For example, `mouse_scroll_lines = 1` moves one step per event.

Agreed scope:

- Keep the default at `3` to preserve current behavior.
- Treat `0` as `1` and show a config warning.
- Apply the setting only where mouse capture is enabled. Logs and document views release mouse capture and use terminal-generated arrow keys, so this change does not control their scroll speed.
- Keep click selection and header sorting available.
- Document the setting and its scope in `docs/configuration.md` and `docs/features.md`.
- Test the default, custom values, zero handling, both scroll directions, list boundaries, and config reload through the public input path.

Discussion: https://github.com/nklmilojevic/sofka/discussions/416

The maintainer approved opening this issue with the scope above.

Before the pull request is ready, run `just check`.

ProbstenHias:
## Summary

Every mouse wheel notch currently moves the selection three rows (hard-coded in `src/app/mouse.rs`). On high-resolution mice and trackpads that overshoots, and there was no way to tune it.

This adds a `mouse_scroll_lines` option to the config:

```toml
mouse_scroll_lines = 3
```

- Default stays `3`, so behaviour is unchanged for existing users.
- Values below `1` are treated as `1`.
- Applied at startup, on config reload, and on context switch, in the same places `remember_sort` is handled.

Proposed in discussion #416. Opened as a draft so we can settle the naming/design there first.

## Changes

- `src/config.rs`: new `mouse_scroll_lines: Option<u16>`
- `src/app/mod.rs`, `src/main.rs`, `src/app/actions.rs`, `src/app/pickers.rs`: carry the value on `App` and populate it wherever config is loaded
- `src/app/mouse.rs`: `wheel()` uses the configured count
- `src/app/tests.rs`: `mouse_scroll_lines_sets_rows_per_wheel_notch` covers 1, 0 (clamped) and unset (back to 3) via `reload_config` + `handle_mouse`
- `docs/configuration.md`, `docs/features.md`: document the option

## Testing

- `rustfmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked` all pass locally (1179 tests).

Closes #424

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -23,6 +23,7 @@ hide_header = false # true hides the header and logo
 compact_mode = false # true starts with a one-line header and no footer
 mouse             = true   # false keeps the terminal's native mouse behavior
                            # (text selection) instead of scroll/click/sort
+mouse_scroll_lines = 3     # rows moved per wheel notch; 1 for line-by-line
 terminal_title = true # false disables terminal title changes
 remember_sort     = true   # save and restore sort choices per resource kind
                            # false makes sort changes temporary
@@ -54,6 +55,12 @@ namespace, context, and live status, with the footer hidden. The default is
 overrides are resolved at startup; `:reload` and subsequent context switches do
 not reset the current compact mode. With `hide_header = true`, the compact
 header is hidden too. Command and filter input still appears when needed.
+
+`mouse_scroll_lines` sets how many rows one mouse wheel notch moves in every
+view (table, logs, documents, pickers). The default is `3`. Set it to `1` if
+your terminal or trackpad already sends several wheel events per gesture and
+scrolling overshoots. `0` behaves like `1`. The option supports cluster and
+context overrides and `:reload`.
 
 `remember_sort` is enabled by default. Set it to `false` to stop saving and
 restoring sort choices from `S`, `I`, and column header clicks. Existing saved
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -60,7 +60,7 @@ header is hidden too. Command and filter input still appears when needed.
 event triggers in views with mouse capture enabled (tables and pickers). The
 default is `3`. Set it to `1` if your terminal or trackpad already sends several
 wheel events per gesture and scrolling overshoots. `0` is treated as `1` and
-reported as a config warning. Logs and document views release mouse capture and
+values above `100` are capped at `100`, each reported as a config warning. Logs and document views release mouse capture and
 receive terminal-generated arrow keys instead, so this setting does not control
 their scroll speed. The option supports cluster and context overrides and
 `:reload`.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -23,7 +23,7 @@ hide_header = false # true hides the header and logo
 compact_mode = false # true starts with a one-line header and no footer
 mouse             = true   # false keeps the terminal's native mouse behavior
                            # (text selection) instead of scroll/click/sort
-mouse_scroll_lines = 3     # rows moved per wheel notch; 1 for line-by-line
+mouse_scroll_lines = 3     # steps per wheel event in captured views; 1 for one step
 terminal_title = true # false disables terminal title changes
 remember_sort     = true   # save and restore sort choices per resource kind
                            # false makes sort changes temporary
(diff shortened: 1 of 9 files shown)


docs/configuration.md
```diff
@@ -55,6 +56,12 @@ overrides are resolved at startup; `:reload` and subsequent context switches do
 not reset the current compact mode. With `hide_header = true`, the compact
 header is hidden too. Command and filter input still appears when needed.
 
+`mouse_scroll_lines` sets how many rows one mouse wheel notch moves in every
+view (table, logs, documents, pickers). The default is `3`. Set it to `1` if
```
nklmilojevic: Please limit this description to views with mouse capture enabled. Logs and document views release mouse capture and receive terminal-generated arrow keys, which bypass `wheel()`. This setting does not control their scroll speed. Update `docs/features.md` too, and describe the count as steps per received wheel event, not per physical notch. This matches the scope in #424.

src/app/actions.rs
```diff
@@ -1451,6 +1451,7 @@ impl App {
         self.user_aliases = resolved.config.aliases;
         self.namespace_favorites = resolved.config.favorite_namespaces;
         self.remember_sort = resolved.config.remember_sort.unwrap_or(true);
+        self.mouse_scroll_lines = resolved.config.mouse_scroll_lines.unwrap_or(3).max(1);
```
nklmilojevic: Please add a config warning when `mouse_scroll_lines = 0` is converted to `1`, as required by #424. The current code converts it silently. Make sure the warning is available at startup, reload, and context switch, and add a test that checks the warning.

nklmilojevic (changes requested):
Please address the two inline comments to match the scope in #424. The config value is applied at startup, reload, and context switch. `just check` passed at 5070ea7.

docs/configuration.md
ProbstenHias: Fixed in 0b7707f. `docs/configuration.md` now describes the count as navigation steps per received wheel event in views with mouse capture (tables and pickers), and states that logs and document views release capture and use terminal-generated arrow keys, so the setting does not control their speed. `docs/features.md` says the same in the mouse support entry, and the `0` handling is documented as a config warning.

src/app/actions.rs
ProbstenHias: Fixed in 0b7707f. Added `config::mouse_scroll_lines(value, &mut warnings)` which returns 3 when unset and pushes `mouse_scroll_lines: 0 is not allowed — using 1` for `0`. It is used in `main.rs` (startup, into `config_warnings`), `reload_config` (into `warnings`), and `apply_context_switch` (into the vector that becomes `config_warnings`). The existing test now asserts the warning after a reload with `0` and that it clears when the key is removed.

nklmilojevic (changes requested):
Please also address Greptile findings. Thanks!

ProbstenHias:
Greptile findings addressed in 8e5e571:

- **Wheel actions cross overlay boundaries**: `wheel()` now resolves the action once and stops iterating as soon as `self.mode` changes, so a wheel event on a confirmation cancels it and leaves the table selection untouched. Covered by `wheel_that_dismisses_a_confirm_does_not_scroll_the_table_beneath` (fails on the previous code). `wheel_and_keyboard_arrows_cancel_default_confirmations` now compares the wheel against one arrow press instead of three, since the wheel stops after cancelling.
- **Scroll count is unbounded**: values above `100` are capped at `100` with a config warning, through the same `config::mouse_scroll_lines` helper as the `0` case, so it applies at startup, reload, and context switch. Test covers `65535` and checks the list boundary still stops the selection. Documented in `docs/configuration.md`.

`rustfmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, and `cargo test --locked` (1180 tests) pass at 8e5e571.

PR #425, opened 2026-09-09
feat: keep the resource view when switching namespace by command

### LINKED ISSUE #422: feat: keep the current resource view when switching namespace by command
## What problem are you trying to solve?

When a user runs `:ns production` from the Deployments view, sofka changes the namespace but opens the Namespaces list. The user must then return to Deployments to see resources in the selected namespace.

## Proposed solution

Make namespace commands use the existing namespace selection behavior:

- `:ns` without an argument opens the Namespaces list.
- `:ns <name>` changes namespace and keeps the current resource type. For example, running `:ns production` from Deployments shows Deployments in `production`.
- From the Namespaces list, follow the existing Enter behavior: return to the previous view if one exists, or show Pods in the selected namespace otherwise.
- Apply the same behavior to aliases that resolve to the Namespaces resource, including `:namespace` and `:namespaces`.
- Treat `all` and `*` as all namespaces on every path, including commands issued from the Namespaces list.
- Keep other cluster-scoped views, such as Nodes, open. Retain the selected namespace for the next namespaced resource view.
- Use the existing namespace switcher's rules for filters and resource ownership. Do not retain a resource ownership scope tied to the old namespace.
- Keep other resource commands, such as `:deploy production`, consistent with their current behavior.

### Acceptance criteria

- Add tests through `handle_key` for namespace commands from a namespaced resource view, the Namespaces list, and another cluster-scoped view.
- Cover namespace aliases, `all`, `*`, the command without an argument, and the previous-view and Pods fallback paths.
- Check that filter and resource ownership behavior matches the existing namespace switcher.
- Update command documentation and help text to explain the namespace command exception.
- Run `just check` before the pull request is ready.

## Alternatives considered

Use the `n` namespace switcher, which already keeps the current resource view, or run an explicit resource command such as `:deploy production`. The proposed change makes direct namespace commands consistent with namespace selection.

## Additional context

Discussion: https://github.com/nklmilojevic/sofka/discussions/410

The maintainer approved creating this issue for the scope above during review of #410. This scope uses the existing Enter behavior from the Namespaces list, including return to a previous view when available.

Related prior implementation: https://github.com/nklmilojevic/sofka/pull/420

nklmilojevic:
`:ns <name>` currently opens the Namespaces list even when the user is viewing another resource. This change keeps the current resource type while selecting the requested namespace. For example, `:ns production` from Deployments shows Deployments in `production`.

Namespace aliases use the same behavior. From the Namespaces list, the command returns to the previous view or opens Pods if no previous view exists. Both paths accept `all` and `*`, use the existing namespace filter rules, and clear stale resource ownership scope. The command without an argument still opens Namespaces. Other cluster-scoped views stay open and retain the namespace selection for later use.

Tests through `handle_key` cover aliases, named and all-namespace selections, Nodes, the command without an argument, previous-view restoration, the Pods fallback, and filter and ownership behavior. Command documentation and help text describe the exception.

Validation: `just check` (format check, clippy with warnings denied, and the full test suite).

Agreed discussion: https://github.com/nklmilojevic/sofka/discussions/410
Closes #422

Replaces #420. This PR retains the original contribution from @Alhiane and adds the corrections, tests, and documentation required by #422. A new PR is needed because maintainer edits are disabled on the original fork branch.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -34,6 +34,14 @@ The full list. For how sofka compares to k9s, see [vs k9s](vs-k9s.md).
   extension API server is down, or it sends an `apiVersion` that is not `v1`.
   Sofka does not load that group. It shows a warning at startup and a flash
   on the first screen. `:info` shows the group and the reason.
+- **Namespace commands**: `:ns <name>` changes namespace and keeps the current
+  resource view, as the namespace switcher does. `:namespace` and `:namespaces`
+  accept the same argument; `all` and `*` select all namespaces. From the
+  Namespaces list, the command returns to the previous view, or opens Pods if
+  there is no previous view. Other cluster-scoped views stay open; the namespace
+  selection applies to the next namespaced resource view. Filters follow the
+  namespace switcher's rules, and resource ownership scope is cleared.
+  Without an argument, `:ns` opens the Namespaces list.
 - **Live watch** of any kind through `kube::runtime::watcher`, streamed into an
   in-memory store. Watch requests use uncompressed responses to avoid gzip
   stream errors. List requests retain gzip compression.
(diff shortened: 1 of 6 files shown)


Alhiane:
Thanks for picking this up and building on #420. The `set_namespace_and_return` -> `set_namespace` consolidation is a cleaner fix for the `all`/`*` gap than what I had, and the added test matrix covers it well.

Checked out this branch locally and confirmed all three CI gates pass: `cargo fmt --all -- --check`, `cargo clippy --locked --all-targets -- -D warnings`, and `cargo test --locked` (1188 passed, 0 failed). Looks good from my side.

PR #485, opened 2026-09-10
chore: publish via trusted publishing and attest release binaries

Tony-Proum:
The crates.io publish authenticated with a long-lived registry token, so the registry recorded the publish as coming from a user account with no evidence tying it to this workflow: a publish from anywhere else, or from anyone holding the token, was indistinguishable from a CI publish. Trusted Publishing exchanges a short-lived OIDC token instead, letting crates.io verify the repository and workflow that asked to publish.

Release binaries had no signature or attestation, and SHA256SUMS is an unsigned file stored beside the tarballs it covers, so whoever could replace an asset could regenerate the checksums with it. Attesting the tarballs binds each digest to this workflow and commit in a signed statement held outside the release, which `gh attestation verify` checks.

Requires a Trusted Publisher registered on crates.io and a `release` environment on the repository before the next release.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -115,6 +115,8 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write # gh release upload
+      id-token: write # Sigstore signing for the provenance attestation
+      attestations: write # record the attestation against the repo
     env:
       GH_TOKEN: ${{ github.token }}
       GH_REPO: ${{ github.repository }}
@@ -129,6 +131,16 @@ jobs:
       - name: Generate checksums
         working-directory: dist
         run: shasum -a 256 -- *.tar.gz > SHA256SUMS
+
+      # Binds every tarball's digest to this workflow, at this commit, in a
+      # Sigstore-signed statement GitHub stores outside the release. Release
+      # assets stay mutable and SHA256SUMS is only as trustworthy as whoever
+      # can rewrite it; `gh attestation verify` is what actually detects a
+      # binary swapped in after the fact.
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@4d101475d8b20a2381f78447822ac1eab6504dd8 # v4.2.2
+        with:
+          subject-path: dist/*.tar.gz
 
       - name: Upload release assets
         run: gh release upload "$RELEASE_TAG" dist/*.tar.gz dist/SHA256SUMS --clobber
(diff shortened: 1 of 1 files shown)


Tony-Proum:
@nklmilojevic I was really please discovering this repository and the tool itself, I was bothered by some points regarding security so I asked claude to help me fix them on your repository. Mostly regarding potential supply chain attacks that are common those days. 

⚠️  before merging this PR, you'll have to do some operations on crate.io in order to validate your / this repository identity and be able to validate each new crate that you push.

Here is more explaination from claude, you'll find the relevant part in the `Action required from you before merging` section

---

Harden the release pipeline: Trusted Publishing + build provenance

Why

A security review of the release pipeline turned up two supply-chain gaps. Neither is a bug in sofka itself — the application code came back clean, and the workflows are already well-hardened (SHA-pinned actions, persist-credentials: false, minimal permissions). These are both about what happens after a build.

1. Nothing proves the crates.io publish came from CI. The crates-publish job authenticates with a long-lived CARGO_REGISTRY_TOKEN. The crates.io API confirms "trustpub_data": null for every published version — the registry records the publish as coming from a user account, not from this workflow. A publish from a laptop, or from anyone who obtained that token, would be indistinguishable in the registry metadata.

2. Release binaries can be swapped after publication. GitHub release assets are mutable, and SHA256SUMS is an unsigned file sitting next to the tarballs — anyone able to replace a binary can regenerate the checksums in the same action. It protects against a corrupted download, not against an attacker. There are currently no build attestations (GET /repos/nklmilojevic/sofka/attestations/… returns 404), so a substituted binary is undetectable by a downloader.

What this PR changes

Only .github/workflows/release.yaml:

- upload-assets now calls actions/attest-build-provenance on the tarballs, producing a Sigstore-signed statement binding each artifact's digest to this workflow and commit. Users can then run gh attestation verify <tarball> --repo nklmilojevic/sofka. Adds id-token: write and attestations: write to that job.
- crates-publish now uses crates.io Trusted Publishing via rust-lang/crates-io-auth-action, exchanging a short-lived OIDC token for a scoped registry token. Adds environment: release and id-token: write, and drops the CARGO_REGISTRY_TOKEN secret reference.

Both new actions are pinned by commit SHA, matching the existing convention.

Action required from you before merging

This cannot land as a code change alone. Two things need to be configured by someone with the right access, and the publish job will fail without them:

1. Register the Trusted Publisher on crates.io. As an owner of the sofka crate: crate page → Settings → Trusted Publishing → Add. The values must match the workflow exactly:

- Repository owner: nklmilojevic
- Repository name: sofka
- Workflow filename: release.yaml
- Environment: release

2. Create the release environment in GitHub repo Settings → Environments. The name must match both the workflow and the crates.io registration. Optionally add protection rules — this is also a good place to restrict who can trigger a publish.

Suggested order

1. Do the two configuration steps above.
2. Merge this PR.
3. Cut a release as usual and confirm crates-publish succeeds.
4. Delete the CARGO_REGISTRY_TOKEN secret and revoke the token on crates.io.

Step 4 is where the security benefit actually lands. Until that token is revoked, it remains a stealable credential and the change is only half-done. I'd suggest keeping it in place until step 3 passes, so a failed cutover doesn

Worth noting: crates-publish runs after upload-assets, so if autheease with binaries uploaded but nothing published to crates.io.
Recoverable, but easier to avoid by doing the setup first.

🤖 Generated with Claude Code

---

Issue #502, opened 2026-09-11
feat: add an official reviewed plugin catalog and management CLI


nklmilojevic:
## Problem and approval

Users can load local plugin packages, but sofka has no standard way to find, install, track, or remove reviewed packages. Manual copying makes versions and updates difficult to manage.

Agreed discussion: https://github.com/nklmilojevic/sofka/discussions/501. The maintainer approved the proposal in the discussion and requested this specification.

Official repository: https://github.com/nklmilojevic/sofka-plugins. The repository starts with a placeholder README. This issue defines the implementation work in both repositories.

## Outcome

Provide an official, reviewed plugin catalog and a cluster-independent `sofka plugin` CLI. Reuse the existing `plugin.toml` package format, validation, execution controls, and reload behavior. Keep external adapters outside the sofka application repository.

“Reviewed” means a maintainer reviewed the published package version. It is not a guarantee that the adapter or its external tools have no defects. Plugins still run with the user's permissions and inherited environment. The catalog does not provide a process sandbox.

## Initial scope

- One official catalog: `nklmilojevic/sofka-plugins`.
- Search, description, installation of one or more packages, installed-package listing, and removal.
- Explicit package versions, compatibility checks, checksums, and installation records.
- Repeatable publication through pull requests and CI.
- Documentation and tests for both the catalog and the client.
- At least one working package with test fixtures to verify the complete workflow before release. Select the first integration during implementation; Popeye and Trivy are candidates, not promises that both ship initially.

Do not add arbitrary remote catalogs, automatic updates, automatic external-tool installation, installation scripts, a TUI catalog browser, or a dependency solver in this first version. Existing manually installed packages remain supported. Do not move bundled plugins out of the application as part of this change.

## CLI contract

```text
sofka plugin search [QUERY]
sofka plugin describe ID[@VERSION]
sofka plugin install ID[@VERSION] [ID[@VERSION] ...]
sofka plugin list
sofka plugin remove ID [ID ...]
```

- Commands do not open the TUI, read cluster credentials, connect to Kubernetes, or execute adapters. They work without a kubeconfig or TTY.
- Preserve existing resource navigation, `sofka info`, `--validate-plugin`, and `--validate-plugin-report`. Document and test an explicit way to open a resource named `plugin` if the new subcommand reserves that name.
- `search` matches IDs, display names, descriptions, and tags without case sensitivity. Sort by ID. Show compatibility and whether a package is installed. No matches is a successful empty result.
[...]

PR #513, opened 2026-09-11
feat: add an official reviewed plugin catalog and management CLI

vyrti:
## Summary

Adds the catalog plugin management CLI: search, describe, install, update,
list, and remove against a commit-pinned reviewed index, with BLAKE3 digests
and zstd archives.

Implements #502 as agreed in discussion #501. Companion:
nklmilojevic/sofka-plugins#1, which now carries five packages — popeye, trivy,
oha, chaos-kill, and resource-summary.

Rebased onto current `main`.

## Performance

Hashing runs on every install, list, update, and removal; decompression on
every install.

| | throughput | 4 MB package |
| --- | --- | --- |
| SHA-256 | 2082 MB/s | — |
| BLAKE3 (SIMD + rayon) | 8476 MB/s | — |
| gzip (was) | 264 MB/s | 1,831,253 B |
| zstd | 728 MB/s | 1,474,482 B |

sofka already builds a C toolchain and
cmake for `aws-lc-sys`, so neither adds a build requirement.

Also: one HTTPS client per command instead of one per request, which was
reloading the system trust store twice per command (~315 ms); and `search` and
`describe` no longer hash every installed file to compute a flag they never
show.

## Tests

66 tests cover the feature directly — 33 in `plugin_install`, 19 in
`plugin_catalog`, 14 in `plugin_cli` — plus 17 in the plugin loader. Coverage
includes describe, JSON output shape, missing tools, digest mismatch, cache
eviction, unmanaged and modified destinations, batch activation failure,
recovery at every filesystem step, and a managed install reaching the loader
through `:reload` via `handle_key`.

Verified end to end against a real catalog: publish workflow, release assets,
digest verification, and extraction, for all five packages. `:popeye` was then
driven against a live cluster and rendered its scan in the TUI.

## Review rounds

Three rounds of review have landed on top of the original change.

**Install safety.** Removal followed symlinked parent directories; activation
swapped without re-verifying the destination; backups were deleted rather than
retired; a partial removal reported total failure. The extractor now bounds the
whole decompressed stream rather than each entry, which a 2,175-byte archive
carrying 2 MiB of GNU long-name metadata defeats otherwise.

**Archive paths.** A path was validated once and then re-derived for the
installation record, so the record could disagree with the file on disk: a
backslash meant a separator on Windows and a filename on Unix, and an entry
named `.sofka-install.json` overwrote the record itself. The validator now
returns the single spelling used for both. It also accepts the trailing slash
an ordinary TAR directory entry carries — previously rejected, so packages
built by any normal `tar` could not install at all.

**Reporting.** `describe` reported the selected release's withdrawal but never
the installed one's, hiding a withdrawn version from the person running it.
Runtime requirements gained `alternatives`, so a package whose adapter finds
any of several executable names reports all of them instead of gating on one.

## Note for review

The catalog format changed in three ways that #502 does not currently
describe: the `[package]` table in `plugin.toml`, BLAKE3 digests, and zstd
archives. Nothing has been published, so no existing record changes — but the
issue text should be amended before this merges.

The catalog endpoints were pointed at a fork while the publish pipeline was
exercised end to end. They are back on `nklmilojevic/sofka-plugins`.


vyrti:
@nklmilojevic ready for **careful** review

src/plugin_install.rs
```diff
@@ -0,0 +1,2663 @@
...
+    declared: &crate::config::Plugin,
+    published: Option<&crate::plugins::Package>,
+) -> Result<(), String> {
+    let mut differences = Vec::new();
+    let mut compare = |field: &str, manifest: &str, catalog: &str| {
+        if manifest != catalog {
+            differences.push(format!("{field} {manifest:?}, catalog says {catalog:?}"));
+        }
+    };
+    if let Some(published) = published {
```
nklmilojevic: This skips the version check entirely when the archive has no `[package]` table, so the P1 from the earlier round is only fixed for manifests that carry it. A catalog package without `[package]` installs with the catalog version recorded and nothing cross-checked.

Make `[package]` required for catalog installs - refuse the stage when `published` is `None`. Hand-installed packages can still leave it out, that's the loader's concern, not the installer's.

src/plugin_install.rs
```diff
@@ -0,0 +1,2663 @@
...
+    pub modified: bool,
+}
+
+pub struct InstallLock {
+    file: File,
+}
+
+impl InstallLock {
+    pub fn acquire(config: &Path) -> Result<Self, String> {
+        ensure_directory_path(config)?;
```
nklmilojevic: `ensure_directory_path` is applied to the config root here and in `prepare` (line 251), not only to the destination. Anyone with `~/.config/sofka -> ~/dotfiles/sofka` (stow, chezmoi, home-manager) gets "refusing symlinked directory" on every install, update and remove.

The spec only forbids following a symlinked destination. Keep the check on `plugins/<id>` and drop it on the root.

src/plugin_catalog.rs
```diff
@@ -0,0 +1,1543 @@
...
+/// 50 MiB needs roughly 14 Mbit/s to land inside the metadata budget, which is
+/// more than a plugin install can assume. The stall budget, not the total, is
+/// what ends a download that has actually died.
+const ARTIFACT_BUDGET: Budget = Budget {
+    total: Duration::from_secs(900),
+    stall: Duration::from_secs(30),
+};
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
```
nklmilojevic: `deny_unknown_fields` on every index struct means the catalog can never add an optional field under schema 1 without breaking every deployed sofka with "invalid catalog JSON: unknown field". And a future `schema_version: 2` that adds fields fails the same way instead of hitting the "unsupported catalog schema_version" message in `validate`.

Deserialize `{ schema_version }` first with a two-field struct, check it, then parse the rest. Drop `deny_unknown_fields` from `Catalog`, `CatalogPlugin`, `CatalogVersion`, `Artifact` and `RuntimeRequirement`. Keeping it on `CachedCatalog` is fine, we own both sides of that one.

src/plugin_catalog.rs
```diff
@@ -0,0 +1,1543 @@
...
+            "cannot determine config directory: set XDG_CONFIG_HOME or HOME".to_string()
+        })?;
+    Ok(base.join("sofka"))
+}
+
+pub async fn load(offline: bool) -> Result<CatalogSnapshot, String> {
+    if offline {
+        return load_cached(&cache_dir().join("catalog-cache.json"));
+    }
+    let commit_bytes = get(COMMIT_URL, CATALOG_MAX_BYTES).await?;
```
nklmilojevic: Only the rate-limit branch in `get_inner` tells the user about `--offline`. DNS failure, connection refused, TLS error or a timeout come out as "requesting <url>: <hyper error>" with no next step. Spec: "Explain how to retry or explicitly use offline mode."

Wrap both `get` calls in `load` so any network error ends with something like "check the network and retry, or pass --offline to use the cached catalog" - and only mention `--offline` when `catalog-cache.json` actually exists.

src/plugin_catalog.rs
```diff
@@ -0,0 +1,1543 @@
...
+    let catalog = Catalog::parse(&bytes)?;
+    let fetched_at = now();
+    let cached = CachedCatalog {
+        schema_version: 1,
+        commit: commit.sha.clone(),
+        fetched_at,
+        catalog: catalog.clone(),
+    };
+    let cache = serde_json::to_string(&cached).map_err(|e| e.to_string())?;
+    crate::atomicfile::write(&cache_dir().join("catalog-cache.json"), &cache)
```
nklmilojevic: A failed cache write fails the whole command here, after the catalog was already fetched and validated. Read-only `$HOME`, full disk or an unwritable `XDG_CACHE_HOME` makes search/describe/install unusable for no reason.

Warn on stderr and continue with the snapshot.

src/plugin_catalog.rs
```diff
@@ -0,0 +1,1543 @@
...
+    }
+    if offline {
+        return Err(format!(
+            "artifact {} is not cached; run install without --offline once",
+            artifact.blake3
+        ));
+    }
+    let bytes = get_with(&artifact.url, ARTIFACT_MAX_BYTES, ARTIFACT_BUDGET, false).await?;
+    verify_artifact(artifact, &bytes)?;
+    write_bytes(&path, &bytes).map_err(|e| format!("caching artifact: {e}"))?;
```
nklmilojevic: Same for the artifact: the bytes are verified against the catalog digest already, the cache write failing shouldn't abort the install. Since `extract` reads from the returned path, either return the bytes alongside the path or fall back to a temp file when the cache dir isn't writable, and warn.

src/plugin_cli.rs
```diff
@@ -0,0 +1,1418 @@
...
+    license: &'a str,
+    readme: &'a str,
+    sofka: &'a str,
+    platforms: Vec<&'a str>,
+    requirements: &'a [crate::plugin_catalog::RuntimeRequirement],
+    command: &'a str,
+    target: &'a str,
+    output: &'a str,
+    mutating: bool,
+    confirm: bool,
```
nklmilojevic, line 119: The text view prints `confirmation: confirms()` (confirm OR dangerous OR network_load) but the JSON only has raw `confirm`. A script reading `.confirm` for a network-load plugin gets `false` although sofka will prompt. The comment at line 261 says text and JSON can't drift apart - they do here.

Add a `confirmation` field to `Description` populated from `confirms()`, and enumerate the describe keys in `docs/plugins.md` the same way search and list are listed. Right now the docs say "the fields printed by the text view" and that's not quite true (`readme`, `tags`, `status`, `installed_withdrawal_reason` are JSON-only).

nklmilojevic (changes requested):
Careful review done, thanks for the three rounds of hardening! Ran `just check` locally, all green. Went through all three new modules line by line - the safety core holds up (path validation, bounded extraction, staged swap, recovery, trash ownership, HTTPS-only redirects, commit-pinned index). No security or data loss bug found.

A few changes before merge, inline below. The two that matter most are the `[package]` requirement in reconcile and dropping `deny_unknown_fields` on the index structs.

Two more things:

- Please rebase on `main`. Merge-base is before v0.25.6 and #530, the branch is 5 commits behind.
- This ships in 0.26.0 as a minor release, not a patch. So the `sofka` bound in nklmilojevic/sofka-plugins#1 has to be `>=0.26.0`, not `>=0.25.5` - no released sofka understands `[package]` or the CLI. Same for the example in `docs/plugin-authoring.md` here.

I'll amend #502 myself with the `[package]` table, BLAKE3, zstd and `update` so the spec matches what lands.

src/plugin_cli.rs
vyrti: Done in 231e109 and 3bcf168. `Description` now has a `confirmation` field populated from `confirms()`, which is back as a helper over `CatalogVersion` carrying the `App::run_plugin` rule, and both the text and JSON views read that one value. `docs/plugins.md` now enumerates the describe keys the way search and list are listed, with a line saying `confirmation` is true whenever `confirm`, `dangerous`, or `network_load` is.

nklmilojevic:
Thanks for the great work! I will merge this, lets work on the plugins repo and then we can release a new version.

PR #518, opened 2026-09-11
feat: Argo CD Application view and columns

mesutoezdil:
#532


nklmilojevic:
hey @mesutoezdil all of these need first an issue and to agree about the scope before working on them. Will close them until we agree in discussion.

PR #520, opened 2026-09-11
feat: expand what an Argo CD managed resource owns

mesutoezdil:


nklmilojevic:
hey @mesutoezdil all of these need first an issue and to agree about the scope before working on them. Will close them until we agree in discussion.

PR #521, opened 2026-09-11
fix: size the confirm dialog to its own text

elidrissi-amine:
## Problem

`draw_confirm` sizes the popup as a fixed share of the frame — 50% of the width with a 56-column floor — and renders `confirm_label` on one unwrapped line. A label longer than the inner width is truncated mid-word.

The drain prompt reaches 67 characters on a node with an ordinary name, so on `main` today, pressing `D` on a node called `bastion-expert-lizard-1` shows:

```
╭ Confirm ─────────────────────────────────────────────────╮
│  Drain node bastion-expert-lizard-1? Cordon and evict eli│
│y:confirm  esc:cancel                                     │
╰──────────────────────────────────────────────────────────╯
```

"eligible pods." is gone, and the hint line sits flush against the border. The same percentage leaves most of the box empty on a wide terminal.

## Change

Measure the label and the key hints, size the popup to the longest of them (keeping the 56-column floor so a short question still reads as a dialog), and grow it by however many rows the text needs. Anything still too long wraps rather than being cut. The block carries its own horizontal padding, which survives wrapping — the label's leading spaces did not.

Same prompt after:

```
╭ Confirm ─────────────────────────────────────────────────────────────╮
│                                                                      │
│ Drain node bastion-expert-lizard-1? Cordon and evict eligible pods.  │
│                                                                      │
│ y:confirm  esc:cancel                                                │
╰──────────────────────────────────────────────────────────────────────╯
```

A multi-line label is supported too: `confirm_label` is split on newlines, so an action that wants a second line gets one.

## Tests

`confirm_dialog_fits_its_label_at_any_terminal_size` drives `D` through `handle_key` at 80x24, 120x32 and 200x40 and asserts both directions — every word of the label and the hints is on screen, and the dialog is no wider than the label needs. It fails on `main` with `80x24 lost "eligible pods."`.

`cargo fmt --all -- --check`, `cargo clippy --locked --all-targets -- -D warnings` and `cargo test --locked` all pass (1381 tests).

Bug fix, so no discussion per `AGENTS.md`.

### DIFF
--- a/src/app/tests.rs
+++ b/src/app/tests.rs
@@ -5592,6 +5592,65 @@ async fn delete_confirm_cascade_can_cycle() {
     ));
 }
 
+/// A label longer than the dialog's minimum width has to stay readable: the
+/// drain question on a node with an ordinary name is 67 characters, which the
+/// fixed 56-column popup truncated mid-word.
+#[tokio::test]
+async fn confirm_dialog_fits_its_label_at_any_terminal_size() {
+    use ratatui::Terminal;
+    use ratatui::backend::TestBackend;
+
+    for (width, height) in [(80u16, 24u16), (120, 32), (200, 40)] {
+        let (mut app, _rx) = test_app();
+        app.switch_kind("nodes");
+        apply(
+            &mut app,
+            json!({"apiVersion": "v1", "kind": "Node",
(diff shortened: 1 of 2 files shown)


elidrissi-amine:
Closing in favour of a PR that carries this together with the drain options it was extracted from.

PR #522, opened 2026-09-11
feat: add the kubectl drain options to the confirm dialog

elidrissi-amine:
## Problem

A drain refuses the whole node when any pod has an `emptyDir` volume or no controller, and silently skips DaemonSet pods. The three `kubectl drain` flags that decide those cases have no equivalent, so a node running a bare pod or an emptyDir pod cannot be drained from sofka at all — `D` cordons it and then reports:

```
drain node-a blocked: default/cache: pod has emptyDir data that eviction would
delete. Node remains cordoned; no pods were evicted from this node
```

The current policy is hardcoded in `drainable_pod` and `drain_blocker`:

| `kubectl drain` flag     | sofka on `main`                                        |
| ------------------------ | ------------------------------------------------------ |
| `--ignore-daemonsets`    | always on — DaemonSet pods are skipped silently        |
| `--delete-emptydir-data` | unavailable — any `emptyDir` pod blocks the whole node |
| `--force`                | unavailable — an uncontrolled pod blocks the whole node |

k9s exposes all of these in its drain form (`internal/view/drain_dialog.go`), which is where the muscle memory comes from.

## Change

`DrainOptions` rides on `ConfirmAction::Drain`, seeded from a new `[drain]` config section, and the dialog toggles it the way delete already toggles force and cascade — two new `confirm`-scope actions, `drain_ignore_daemonsets` and `drain_emptydir_data`, plus the existing `force`:

```
╭ Confirm ─────────────────────────────────────────────────────────────╮
│                                                                      │
│ Drain node bastion-expert-lizard-1? Cordon and evict eligible pods.  │
│ ignore daemonsets: yes  ·  delete emptyDir data: no  ·  force: yes   │
│                                                                      │
│ y:confirm  esc:cancel  i:ignore daemonsets  e:emptyDir data  f:force │
╰──────────────────────────────────────────────────────────────────────╯
```

The dialog lists the three values so the effect is visible before `y`, and each blocker names the option that waives it. `[drain]` sets the defaults the dialog opens with, and they keep the current behaviour — ignore daemonsets on, the other two off — so nothing changes for anyone who does not configure or toggle them.

One semantic worth calling out: **DaemonSet pods are never evicted either way**, because the controller would recreate them on the cordoned node. `ignore_daemonsets` only chooses between skipping them and refusing the node, which is what `kubectl` does without the flag.

Grace period, timeout and `--disable-eviction` are deliberately left out. sofka does not wait for pods to terminate (it reports `drain requested` once the API accepts the evictions), and `docs/safety.md` promises drain never falls back to direct deletion, which `--disable-eviction` would break.

## Commits

1. `fix: size the confirm dialog to its own text` — a prerequisite, and a bug on `main` on its own: the dialog took a fixed share of the frame and rendered the label unwrapped, so the 67-character drain prompt for a node with an ordinary name was truncated to "Cordon and evict eli". It now measures its text, keeps the 56-column floor for short questions, wraps the rest, and supports a multi-line label — which the options line needs. Offered standalone in #521; folded in here since the feature depends on it.
2. `feat: add the kubectl drain options to the confirm dialog`.

## Tests

Behaviour is driven through `handle_key` throughout:

- `drain_toggles_waive_the_emptydir_and_force_blockers` — each toggle lets the previously blocked pod be evicted, asserted against the recorded API calls.
- `drain_ignore_daemonsets_toggle_refuses_instead_of_skipping` — a DaemonSet pod is never evicted, and turning the option off refuses the node.
- `drain_dialog_text_follows_the_option_toggles` and `drain_dialog_shows_all_options_without_clipping` — the dialog text tracks the toggles and renders in full at three terminal sizes.
- `drain_config_defaults_seed_the_dialog` — `[drain]` supplies the starting point.
- `confirm_dialog_fits_its_label_at_any_terminal_size` — the sizing fix; fails on `main` with `80x24 lost "eligible pods."`.

Docs updated in `safety.md` (new option table; the old "Drain has no override for unmanaged pods or `emptyDir` data loss" line is no longer true), plus `keys.md`, `keybindings.md`, `configuration.md` and `features.md`.

`cargo fmt --all -- --check`, `cargo clippy --locked --all-targets -- -D warnings` and `cargo test --locked` pass, and each commit builds and passes the suite on its own (1381 then 1386 tests).

## Process note

`AGENTS.md` asks for a discussion and maintainer agreement before a feature PR, and I do not have one for this — happy to open one and park this until you have weighed in on the scope, or to close it if you would rather shape the option surface differently. The implementation is here because it was already written and running against our clusters.


nklmilojevic:
Please read the pinned issue on how to contribute to the project. Thanks.

elidrissi-amine:
Pushed a revision addressing the review, and a note on the red CI.

**CI is failing before any job runs, for reasons unrelated to this branch.** `ci.yaml` on `main` points the home-manager job at `$/.github/actions/setup-nix`, which is not a valid `uses` value, so the workflow never schedules: the run for this PR completed with `total_count: 0` jobs and "This run likely failed because of a workflow file issue". The same happens on #521 and #518. #524 fixes that one line; once it lands, the rust and coverage jobs here will actually report something. Locally `cargo fmt --all -- --check`, `cargo clippy --locked --all-targets -- -D warnings` and `cargo test --locked` pass, and each of the two commits passes the suite on its own (1382 then 1387).

**Non-controlling DaemonSet references** (P1, `helpers.rs`): fixed, and thanks — this was a real hole. `daemonset_pod` matched any owner reference of kind DaemonSet, so a pod carrying a stale or non-controlling one was skipped as the DaemonSet's business and never reached the unmanaged-pod rule, letting the drain report success while leaving it on the cordoned node. It now reads only the controlling owner, the way `kubectl` does via `GetControllerOf`. That made an existing fixture in `drainable_pod_skips_daemonset_mirror_and_completed_pods` unrealistic (its DaemonSet reference had no `controller: true`, which real DaemonSet pods always carry), so the fixture now sets it, and a new case pins the distinction: a non-controlling reference is drainable and blocks on "no controller".

**Display width** (P2, `ui.rs`): fixed. Both the width and the row count now use `UnicodeWidthStr::width` instead of scalar counts, so a double-width label or key hint no longer sizes the dialog for half the columns it occupies. `confirm_dialog_measures_double_width_labels` renders a CJK label and asserts its tail survives.

**Orphaned pods and `force`** (P1, `helpers.rs`): not changed, documentation corrected instead. The check is metadata-only, so a pod whose controller has since been deleted still counts as managed — `kubectl` resolves the reference and this does not. That predates this PR (it is the existing `drain_blocker` behaviour), and resolving owners would mean an API lookup per owner reference on every drain, which felt like a separate change to propose rather than smuggle in here. The `[drain]` doc comment claimed "pods left by a deleted owner", which overstated it; it now says the reference is read, not resolved, and points out the difference from `kubectl`. Happy to do the lookup in a follow-up if you would like drain to match `kubectl` exactly there.

elidrissi-amine:
Understood, and sorry for the wrong order — opened https://github.com/nklmilojevic/sofka/discussions/525 for the scope. Will leave this closed and put a PR up linked to that discussion if you agree it is worth having.

PR #537, opened 2026-09-12
feat: Argo CD Application columns and :argocd view

mesutoezdil:
part of 532


nklmilojevic:
Seems good, merging, lets continue!

PR #548, opened 2026-09-12
fix: keep the all-namespaces scope out of the recent namespaces

MichaelKlank:
## What

`note_recent_namespace` guarded against `<all>` only, while its callers pass
the namespace as the user typed it. `:pods all` and `:pods *` were therefore
stored as recent namespaces: they take slots in the bounded per-context history
(`MAX_RECENT_NAMESPACES`, 8) and push real namespaces out of it.

Normalising inside the guard covers every caller — `set_namespace`,
`switch_kind_ns` and `apply_resource_query` all route through this one
function.

## Test

`the_all_namespaces_scope_is_not_a_recent_namespace` drives it through
`handle_key`: pick a real namespace, then run `:pods all`, `:pods *` and
`:pods <all>`, and assert the active scope switched to all-namespaces while
none of the three spellings was recorded as a recent namespace and the real one
still is. Scope of the test: the active scope and the stored recent history. It
fails on main and passes with the fix.

`just check` passes (fmt-check, clippy `-D warnings`, 1548 tests).

## Note

No discussion linked — `AGENTS.md` exempts bug fixes.


nklmilojevic (changes requested):
The fix addresses a small defect in the recent namespace history, but the description and new code comment claim a user-visible failure that the code and test do not support.

`filtered_namespaces` includes a recent entry only when `available.contains(r)` is true. Recording `all` or `*` therefore does not, by itself, add the claimed broken rows to the namespace picker. The supported effect is that these values occupy slots in the bounded recent history and can remove real namespaces from that history.

Please make these changes:

1. Correct the PR description and commit message to state the supported effect. Remove the claim that the picker offers entries that select nothing. If you believe that failure exists, provide exact reproduction steps and a test that checks the picker rows and selection result.
2. Remove the new explanatory comments from both `src/app/pickers.rs` and `src/app/tests.rs`. The implementation and test are clear without them, and the implementation comment repeats the unsupported claim.
3. Keep the regression test, but describe its scope correctly. It checks the active scope and stored recent history. It does not demonstrate a picker failure.

Please verify reported user-visible failures against the actual code path before stating them as facts. The history fix is reasonable, but its description must match the evidence.

MichaelKlank:
All three points done in 47031a1.

You are right about the picker, and I checked it rather than taking the correction on faith: `filtered_namespaces` only adds a recent when `available.contains(r)`, so `all` and `*` never become picker rows. The claim was wrong and I should have traced that path before writing it as fact.

1. Commit message and PR description now state only the supported effect: the values take slots in the bounded recent history (8 per context) and push real namespaces out of it. The picker claim is gone from both.
2. Both new comments removed, in `src/app/pickers.rs` and `src/app/tests.rs`.
3. Test kept, and its scope is now described as what it checks — the active scope and the stored recent history, not picker rows or selection.

PR #570, opened 2026-09-13
fix: resolve Argo CD destination names against every kubeconfig context

ShmuelOps:
## Problem

On a hub cluster running Argo CD, `:argocd` on an Application that deploys to a spoke by `spec.destination.name` reported

```
these resources live in eks-prod-general, which no kubeconfig context serves
```

even though the kubeconfig had a context for that cluster. A sibling Application targeting `eks-dev-general` resolved fine, which made it look random.

## Root cause

`resolve_destination` checked the name against `context_servers()`, a `HashMap<server, context>`. With two contexts on one cluster (a short alias `prod` next to `eks-prod-general`, both from `aws eks update-kubeconfig --alias`), the map keeps whichever the kubeconfig lists last. `prod` sorts after `eks-prod-general` and wins; `dev` sorts before `eks-dev-general` and loses. So whether Argo's spelling survived depended on alphabetical order.

Separately, the name was only ever compared to context names. EKS kubeconfig entries are named by ARN (`arn:aws:eks:us-east-1:…:cluster/eks-prod-general`) while Argo registers the bare cluster name, so without a hand-named context the lookup could never match.

## Fix

- `Cluster::context_clusters()` returns every `(context, cluster entry)` pair in kubeconfig order, without collapsing duplicates.
- `argocd::context_for_name()` resolves a destination name in three tiers, first hit wins: exact context name → cluster entry name → cluster entry ending in `/<name>` (the ARN tail).
- `classify_destination` takes a `context_for_name: Fn(&str) -> Option<String>` in place of `is_known_context: Fn(&str) -> bool`, so the resolved context (which may be the alias) is what the view shows and what `:ctx` is told to switch to.
- Server-URL destinations are untouched.

## Tests

- New `argocd::tests::destination_name_resolves_by_context_cluster_or_arn_tail` covers all three tiers, the alias-vs-full-name collision in both orders, alias-only kubeconfigs, and that the tail match is not a suffix match.
- `app::tests::argocd_destination_classification` updated for the new closure shape, plus a case where the lookup returns an alias.
- `cargo fmt --all -- --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked` all pass (1634 tests).

Docs: `docs/features.md` Argo CD view paragraph now states the matching rules.

This is a bug fix, so no prior discussion per AGENTS.md. Happy to split the ARN-tail tier into its own PR if you'd rather treat that part as a feature.


src/k8s.rs
```diff
@@ -322,6 +322,51 @@ fn sanitize_server_version(version: &str) -> String {
...
+        for c in &config.contexts {
+            let Some(cluster) = c.context.as_ref().map(|ctx| ctx.cluster.clone()) else {
+                continue;
+            };
+            if let Some(server) = servers.get(cluster.as_str())
+                && !server.is_empty()
+            {
+                index.by_server.insert(server.clone(), c.name.clone());
+            }
+            index.clusters.push((c.name.clone(), cluster));
```
nklmilojevic, line 370: Please only add a context to the name index when its referenced cluster has a nonempty server, while still retaining every valid alias. This push is outside the validation above, so a context referencing a deleted/missing cluster entry (or one without a server) now resolves as a usable destination; the previous server-backed lookup excluded it. For example, a stale context named `eks-prod-general` wins the exact-name tier over a working `prod` context pointing to the EKS ARN, yielding a `:ctx eks-prod-general` suggestion that cannot connect. Please add regression coverage for a dangling cluster reference and a missing/empty server, including a valid alias that must remain eligible.

src/app/tests.rs
```diff
@@ -24171,9 +24171,19 @@ fn argocd_destination_classification() {
...
+        }),
         Destination::Context("in-cluster".into())
     );
+    // A registered name resolves to whatever context the kubeconfig lookup
+    // names for it — the alias, not the Argo spelling.
+    assert_eq!(
+        super::argocd::classify_destination("", "eks-prod-general", here, none, |_| {
+            Some("prod".to_string())
+        }),
+        Destination::Context("prod".into())
```
nklmilojevic, line 24251: Please also cover the changed user-visible behavior through `handle_key`, as required by AGENTS.md. This assertion checks only the classification helper. Add an Argo managed-resource regression that presses Enter with a name resolved to an alias and verifies the resulting message suggests `:ctx prod` (the resolved alias), rather than the Argo spelling or the unresolved-destination message. Keep the helper tests for the lookup tiers.

nklmilojevic (changes requested):
The core duplicate-context fix and the ambiguity guards look sound. Local `just check` passes (formatting, Clippy, 1,644 tests passed; 2 ignored). Please address the stale-context regression and add the key-driven regression coverage noted inline.

src/k8s.rs
ShmuelOps: Done in 31e45a4. `ContextIndex::from_kubeconfig` now admits a context only when its cluster entry exists with a nonempty server, and both the server map and the name list come from that same set — so a stale `eks-prod-general` pointing at a deleted cluster is skipped and the working `prod` alias resolves. Regression `k8s::tests::context_index_skips_contexts_without_a_server` covers a dangling cluster reference, a cluster with no `server`, an empty `server`, and two valid aliases on one entry that both stay eligible.

src/app/tests.rs
ShmuelOps: Added in 31e45a4: `app::tests::argocd_view_enter_names_the_alias_context_for_a_remote_name` drives it through `handle_key` — opens `:argocd` on an Application with `destination.name: eks-prod-general`, presses Enter on a managed resource, and asserts the flash says `switch with :ctx prod` (the alias), not the Argo spelling and not the unresolved message. It runs twice: with only the `prod` alias, and again (via `r`) with a stale `eks-prod-general` context on a deleted cluster also present. To keep it independent of the developer's kubeconfig the view resolves through a `#[cfg(test)]` snapshot override on `App` (`context_index_override`), same spirit as tests setting `all_contexts` directly; happy to shape that differently if you'd prefer. The helper-tier tests stay as they were.

nklmilojevic:
Thanks for the contribution!

Issue #532, opened 2026-09-12
feat: Argo CD read side - Application columns, :argocd view, fleet count, ownership tree


nklmilojevic:
Agreed scope from #412 with @mesutoezdil. Argo CD gets the same read side Flux already has: curated Application columns, an `:argocd` view, a count in `:fleet`, and an ownership tree behind a keypress. Everything comes from the Application CRD, no Argo API or token.

Three PRs, in this order, each mergeable on its own with tests and docs. The closed drafts #518, #519, #520 already follow this split, so rebasing them onto this issue is the expected path.

## Ground rules

- CRD only. Read `argoproj.io/v1alpha1` Applications. Never the Argo CD API, a token, or Argo's cluster Secrets.
- Same shape as Flux. Pure module `src/argocd.rs` (unit tested without a cluster) plus a driver `src/app/argocd.rs`, mirroring `gitops.rs` / `app/gitops.rs`. The view reuses `explain::Finding`, `Level`, `Target` and `ui::draw_findings`. No new rendering code.
- Reads only. Sync and suspend stay on `t`.
- Tests through `handle_key` for every user-visible behaviour. `Cluster::fake()` already registers `argoproj.io/applications`.

### Columns

`columns_for("argoproj.io", "applications")` returns a curated set, six columns like Flux:

| Column      | Source                                                                              |
| ----------- | ----------------------------------------------------------------------------------- |
| `NAME`      |                                                                                     |
| `SYNC`      | `status.sync.status`                                                                |
| `HEALTH`    | `status.health.status`                                                              |
| `REVISION`  | `status.sync.revision`, falling back to `status.operationState.syncResult.revision` |
| `AUTO-SYNC` | three states, see below                                                             |
| `AGE`       |                                                                                     |

- `REVISION` shortens a full 40-char SHA to 7. Chart versions and branch names stay as they are.
- `AUTO-SYNC` is three states, not a boolean:
  - `on`: `spec.syncPolicy.automated` is present
  - `manual`: no `automated` block and no stash annotation. Normal config, not a fault.
  - `suspended`: no `automated` block and `sofka.io/argocd-automated` is set (what `t` → Suspend writes, see `app/helpers.rs`)
- `SYNC` / `HEALTH` sort as text, the way `FLUX_OBJECT_COLUMNS` `READY` does.
- ApplicationSets keep default columns.

### `:argocd` view

[...]

PR #572, opened 2026-09-14
feat: Argo Rollouts — enter drills into pods, t promotes/aborts/retries/pauses/restarts

ShmuelOps:
## Problem

A Rollout (`rollouts.argoproj.io`) row behaves like an unknown CRD: `⏎` opens its YAML, `t` is refused ("suspend/resume only applies to…"), and the header hints `enter yaml`. Driving a canary — promote it, abort it, restart it — means leaving sofka for the Argo Rollouts dashboard or the kubectl plugin, even though sofka already has the `t` action-menu pattern for Argo CD, Flux, and CronJobs.

## What this adds

- **`⏎` on a Rollout drills into its pods** by `spec.selector.matchLabels`, scoped as `rollout/<name>`, exactly like a Deployment.
- **`t` on a Rollout** opens an "Argo Rollouts" menu: Promote, Promote full, Pause, Retry, Abort, Restart. Works on marked rows (bulk) like the other menus.
- **Header hints** on Rollouts read `enter pods` and `t promote/abort/…`.
- Both are guarded by the `argoproj.io` API group (same shape as the CAPI `machinedeployments` arm), so a kind that happens to share the plural keeps its configured drill or YAML.

## Patches (same as `kubectl argo rollouts`)

Taken from the plugin's `promote.go` / `abort.go` / `retry.go` / `restart.go` / `pause.go`:

| Verb | Patch |
|---|---|
| Promote | unpause `spec.paused`; clear `status.pauseConditions`; after an inconclusive analysis under controller pause, also clear `controllerPause` and step on; for a canary mid-analysis with nothing paused, step to the next `currentStepIndex` |
| Promote full | unpause; `status.promoteFull=true` unless `currentPodHash == stableRS` |
| Pause | `spec.paused=true` |
| Retry | `status.abort=false` |
| Abort | `status.abort=true` |
| Restart | `spec.restartAt=<now>` |

Status patches go through the status subresource (`patch_status`); on a 404 (subresource not served) the plugin's unified patch is sent to the object instead, then the `spec` half. Promote GETs the live Rollout first because which fields to clear depends on where it is paused. No plugin binary needed.

## Tests

- `rollout_promote_patches_follow_the_plugin` — the promote decision tree case by case (pause step, `spec.paused` only, mid-analysis stepping incl. the last-step clamp and unset index, inconclusive + controller pause, full with/without a stable hash).
- `rollout_verb_patches_match_the_plugin` — abort / retry / pause / restart bodies.
- `rollout_menu_offers_the_rollout_verbs_and_acts`, `rollout_menu_acts_on_marked_rows` — through `handle_key`, incl. bulk.
- `enter_on_a_rollout_drills_into_its_pods` — through `handle_key`, incl. `esc` back.
- `header_hints_name_the_rollout_drill_and_menu`.
- `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked` (1644) pass.

Docs: `docs/features.md` (new "Argo Rollouts controls" bullet), `docs/keys.md` (enter + `t` rows), and the `?` help string for `t`.

## Note on process

AGENTS.md asks for a Discussion before feature work; this is opened as a PR directly because the change is a straight extension of the existing `t`/drill patterns rather than new surface — but if you'd rather have it go through a Discussion first, say so and I'll convert it. Not verified against a live Rollout beyond the patch bodies matching the plugin source.


nklmilojevic:
Please discuss this in the issue before creating PRs https://github.com/nklmilojevic/sofka/issues/532

PR #573, opened 2026-09-14
feat: navigate from the Argo CD view into the cluster and back

ShmuelOps:
## Problem

On a hub cluster running Argo CD, most Applications deploy elsewhere. The Argo CD view (`:argocd`) shows their state honestly but is a dead end: `⏎` on a managed resource of a remote Application flashes "these resources live in <ctx>; switch with :ctx <ctx>", and the user then switches, picks the kind, picks the namespace, and filters by name themselves — four steps to look at the pod behind a Degraded row, and no way back except doing it all again in reverse. Separately, `⏎` on an Application row opens its YAML (the generic fallback), when the reason to open an Application is its sync/health state, which is what `:argocd` shows.

Follows #570, which made remote destinations resolve to a kubeconfig context in the first place.

## What this adds

Six commits, one feature: getting from the Argo CD view to the objects and back.

1. **Remote jump.** `⏎` on a managed resource whose destination resolved to a context switches to it and, once the connection lands, opens the object as a root view scoped to its name (`metadata.name=` field selector, same as `navigate_to_target`). The kind is resolved against the destination cluster at landing time, so a CRD the hub lacks is fine. Reuses the deferred-navigation pattern of `:pods @ctx ns` — its own `pending_*` slot, same one-at-a-time rules, cleared on a failed switch. Headings and unresolved destinations keep today's flash.
2. **`⏎` on an Application / ApplicationSet row opens the Argo CD view**, not its YAML; `y` still has the YAML. Guarded by the `argoproj.io` group like the CAPI `machinedeployments` arm.
3. **`esc` comes back.** The jump records the hub context and the object the view was opened for. `esc` at the jumped view's root switches back and reopens the Argo CD view on that object, with the table underneath set to that kind. Moving to another root view in the remote cluster drops the way back, so `esc` there stays an ordinary root `esc`.
4. **Header hints** on Applications read `enter argocd` / `t suspend/sync`.
5. **`→` markers** on remote managed-resource rows (they jump now) — `draw_findings` takes the jump predicate from the view instead of the finding, Explain/GitOps unchanged.
6. **Managed resources as a workspace.** `⏎` on the "Managed resources" heading opens one live table per kind in the destination cluster, each filtered `name=a || name=b` to the objects Argo listed (by name, not by the instance label: with annotation tracking Argo sets no label, and a Helm-rendered `app.kubernetes.io/instance` carries the release name). `Tab` cycles kinds; `esc` at a table's root returns to the Argo CD view. Goes through the existing workspace path (`open_workspace`), which gains an optional return.

No new keys, no new config. `docs/features.md` and `docs/keys.md` updated in the same commits.

## Tests

All user-visible behaviour is driven through `handle_key`:

- `argocd_view_enter_on_a_remote_resource_opens_it_in_its_context` — Enter → switch armed → `land_context` → services view scoped to the name → `esc` → switch back → Argo view reopened.
- `argocd_view_enter_on_a_remote_heading_does_not_switch`, `argocd_remote_jump_return_is_dropped_by_a_new_root_view`.
- `enter_on_an_argocd_application_opens_the_argocd_view`.
- `argocd_managed_resources_heading_opens_a_workspace_and_esc_returns` (local), `argocd_managed_resources_workspace_follows_a_remote_destination` (remote), `managed_workspace_views_group_by_kind` (pure grouping/filter/namespace rules).
- `argocd_view_enter_switches_to_the_alias_context_for_a_remote_name` — the #570 regression, adapted: Enter now switches to the alias.
- `header_hints_name_the_argocd_drill_and_menu`.
- `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked` (1646) pass.

## Note on process

AGENTS.md asks for a Discussion before feature work. This is opened as a PR directly because every piece extends an existing pattern (deferred navigation, drills, workspaces, `→` markers) rather than adding new surface — happy to convert it to a Discussion first, or split it (e.g. commits 1–3 vs 6), if you prefer.


nklmilojevic:
Please discuss this in the issue before creating PRs https://github.com/nklmilojevic/sofka/issues/532

PR #630, opened 2026-09-17
feat: jump from the Argo CD view into a remote managed resource, and back

ShmuelOps:
Items 1 and 2 of the proposal in nklmilojevic/sofka#532, as agreed with @mesutoezdil, plus the `→` marker that makes the jump discoverable. Three commits, one feature: getting from the Argo CD view to a remote object and back.

## Problem

Since #570 a remote destination resolves to a kubeconfig context, but `⏎` on a managed resource of a remote Application only reports where it lives and tells the user to `:ctx` there by hand, then find it again — four steps to look at the pod behind a Degraded row, and the same in reverse to come back.

## Change

1. **`⏎` on a remote managed resource jumps to it.** Switch to the resolved context and, once the connection lands, open the object as a root view scoped to its name (`metadata.name=` field selector, same as `navigate_to_target`). The kind is resolved against the destination cluster at landing time, since the hub may not know the CRD; a kind the destination does not know either falls back to pods with a warning. The deferred target lives in its own slot next to the pending query / bookmark / workspace and follows the same one-at-a-time rules, cleared on a failed switch.

   The #532 rule "on a remote destination `enter` flashes where the object lives instead of opening a same-named object here" guarded against opening the wrong cluster's object when destinations could not be resolved. With #570 they resolve to a context, so the guard becomes the jump; an *unresolved* destination still flashes, and a heading or source line under a remote destination still has nothing to open and says so.

2. **`esc` comes back.** The jump records the hub context and the object the view was opened for while the view it opened is on screen. `esc` at that view's root switches back and, once the switch lands, reopens the Argo CD view on the same object, with the table underneath set to that kind. `Tab` through the default kinds in the remote cluster keeps the way back (still browsing where the jump landed); a `:resource` query, bookmark or workspace drops it, so `esc` there is an ordinary root `esc`. The pods fallback keeps it too.

3. **`→` on the rows that now jump.** `draw_findings` takes the jump predicate from the view instead of reading the finding's target; the Argo CD view marks a managed-resource row whenever its destination resolved to a context. Explain and GitOps keep the target-based rule.

No new keys, no new config. Docs: `docs/features.md` Argo CD view paragraph.

## Tests (through `handle_key`)

- `argocd_view_enter_on_a_remote_resource_opens_it_in_its_context` — Enter → `→` on the row → switch armed → `land_context` → services scoped to the name → `esc` → switch back → `Mode::Argocd` with the same title.
- `argocd_view_enter_on_a_remote_heading_does_not_switch`, `argocd_remote_jump_return_is_dropped_by_a_new_root_view`, `argocd_remote_jump_return_survives_tab_cycling`, `argocd_remote_jump_unknown_kind_falls_back_to_pods_and_esc_returns`.
- `argocd_view_enter_switches_to_the_alias_context_for_a_remote_name` — the #570 regression adapted: Enter switches to the alias, not Argo's spelling.
- `argocd_view_will_not_jump_into_a_remote_destination` (unresolved) unchanged.

`cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked` (1705) pass.


ShmuelOps:
Addressed the three review findings in cf1e2b2 (amended into the single commit):

- **Summary row.** `managed_row_ordinal` now returns `None` past `MAX_LISTED`, so the "… and N more" line carries no `→` and `⏎` on it reports where the objects live instead of opening the first hidden resource. Test: `argocd_view_summary_row_is_not_a_jump` (51 resources, remote destination).
- **Current context by name.** `classify_destination` takes the connected context and maps a resolved context equal to it — by name or by server — to `Destination::Current`, so rows get ordinary local targets and `⏎` opens the object here rather than arming a switch that never lands. Tests: two new cases in `argocd_destination_classification`, plus `argocd_destination_resolving_to_the_current_context_is_local` through `handle_key`.
- **Failed switch back.** On a failed `ContextSwitched` the pending return is handed back to `argocd_return` instead of dropped, since the remote table is still on screen; `esc` can retry. Test: `argocd_remote_jump_failed_return_keeps_the_way_back`.

fmt / clippy `-D warnings` / `cargo test --locked` (1708) pass.

ShmuelOps:
Follow-up in d9048fa for the locality finding: a resolved context now counts as local only when its kubeconfig server equals the connected `cluster_url` (normalized) — never by context name. `ContextIndex` gained `server_by_context` for that (every admitted context, no alias collapse), and `classify_destination` takes a `server_for_context` lookup. The residual case — the connected context's *name* now pointing at another server — stays `Context`, and `⏎` says "kubeconfig context X no longer points at the connected cluster; reconnect with :ctx X" instead of arming a no-op switch. Covered in `argocd_destination_classification` (name→same server, name→other server, server→same server) and in `argocd_destination_resolving_to_the_current_context_is_local` through `handle_key`. 1708 tests pass.

src/app/argocd.rs
```diff
@@ -270,12 +293,137 @@ impl App {
...
+    }
+
+    /// Whether `⏎` on row `index` goes somewhere: the row's own target, or a
+    /// managed resource of an Application deploying to a cluster some
+    /// kubeconfig context serves, which jumps through that context.
+    pub fn argocd_row_jumps(&self, index: usize) -> bool {
+        self.argocd_items
+            .get(index)
+            .is_some_and(|f| f.target.is_some())
+            || (matches!(self.argocd_destination, Destination::Context(_))
```
mesutoezdil, line 417: empty resource list still marks none reported with an arrow

src/app/argocd.rs
```diff
@@ -270,12 +293,137 @@ impl App {
...
+        let Destination::Context(context) = self.argocd_destination.clone() else {
+            self.flash_warn(&self.no_jump_reason());
+            return;
+        };
+        // Same name as the connected context, yet not classified local: the
+        // kubeconfig now points that name at another server. A switch would
+        // be a no-op, so say what happened instead of dropping the jump.
+        if context == self.cluster.context && self.cluster.connected {
+            self.flash_warn(&format!(
+                "kubeconfig context {context} no longer points at the connected cluster; reconnect with :ctx {context}"
```
mesutoezdil, line 311: switch_context early returns, so this ctx advice does nothing

src/app/argocd.rs
```diff
@@ -12,6 +12,23 @@ type KindPlurals = HashMap<(String, String), String>;
...
+}
+
+/// The Argo CD view a remote jump left: the context it was on and the object
+/// it was opened for (an Application, or something Argo manages), so the same
+/// view can be reopened after switching back.
+#[derive(Clone)]
+pub(super) struct ArgocdReturn {
+    pub context: String,
+    pub source: DynamicObject,
+    pub kind: crate::k8s::Kind,
```
mesutoezdil, line 29: esc drops the row you left, managed_row_at already exists

src/ui.rs
```diff
@@ -4558,13 +4558,17 @@ fn draw_fleet(frame: &mut Frame, app: &mut App, area: Rect) {
...
+/// jumps its findings do not carry. Shows `empty_msg` while the findings are
+/// still gathering.
+#[allow(clippy::too_many_arguments)]
 fn draw_findings(
     frame: &mut Frame,
     show_scrollbars: bool,
     area: Rect,
     title: String,
     findings: &[crate::explain::Finding],
+    jumps: &dyn Fn(usize, &crate::explain::Finding) -> bool,
```
mesutoezdil, line 4571: arrow marker is item 5, outside the approved scope

src/ui.rs
```diff
@@ -4641,12 +4647,16 @@ fn draw_argocd(frame: &mut Frame, app: &mut App, area: Rect) {
     // The expansion is only discoverable from the title, the way the adjacent
     // view advertises the same key.
     let title = format!(" {} (c discover children) ", app.argocd_title);
+    let jumps: Vec<bool> = (0..app.argocd_items.len())
```
mesutoezdil, line 4650: no testbackend test covers this arrow rewiring at all

src/app/argocd.rs
```diff
@@ -270,12 +293,137 @@ impl App {
...
+        self.pending_bookmark = None;
+        self.pending_workspace = None;
+        self.pending_argocd_target = None;
+        self.pending_argocd_return = Some(back);
+    }
+
+    /// The landing half of [`Self::return_to_argocd`]: the table underneath
+    /// is the kind the view was opened from, so `esc` out of the view lands
+    /// somewhere sensible, and the view itself re-reads its object.
+    pub(super) fn reopen_argocd(&mut self, back: ArgocdReturn) {
```
mesutoezdil, line 361: reopen_argocd should restore the managed row, not row zero

mesutoezdil:
body says three commits and 1705 tests, both stale

mesutoezdil:
same commit is also open at shmuelops/sofka#1

src/ui.rs
mesutoezdil: reverting the arrow in draw_argocd still passes 1708 tests

src/app/argocd.rs
mesutoezdil: tested live: left row 45, esc returns to row 0

mesutoezdil:
live service-east to apps-dev-east: jump works, return loses place

mesutoezdil (changes requested):
two bugs, one unfinished round trip, one unapproved item.

none reported rows advertise a jump they cannot make.
the drifted kubeconfig flash advises a ctx switch that early returns.
esc comes back to row zero, verified live at 45 rows deep.
the arrow marker is item 5, and reverting it still passes every test.

nr1 and nr2 otherwise work on a real hub and spoke.

nklmilojevic:
I'm also not a fan of that many comments in the code. Which model are you using? Seems that it was low quality output from the get go with so many errors that greptile found from the start.

src/app/argocd.rs
```diff
@@ -270,12 +293,137 @@ impl App {
...
+            return;
+        };
+        // Same name as the connected context, yet not classified local: the
+        // kubeconfig now points that name at another server. A switch would
+        // be a no-op, so say what happened instead of dropping the jump.
+        if context == self.cluster.context && self.cluster.connected {
+            self.flash_warn(&format!(
+                "kubeconfig context {context} no longer points at the connected cluster; reconnect with :ctx {context}"
+            ));
+            return;
```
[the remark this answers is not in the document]
ShmuelOps: Fixed in 4400f2e: the jump goes through `switch_context_inner(context, true)` — the reload path upstream added — so a destination carrying the connected context's name reconnects through it instead of returning early. The advice flash is gone. `argocd_destination_resolving_to_the_current_context_is_local` covers it: same name pointing at another server → Enter arms a switch to that name and lands on the object.

src/app/argocd.rs
ShmuelOps: Fixed in 4400f2e. The arrow is gone entirely (item 5, see below), and `managed_row_ordinal` is bounded by both `MAX_LISTED` and `argocd_resources.len()`, so "none reported" and "… and N more" map to no resource. Test: `argocd_view_none_reported_row_is_not_a_jump`.

src/app/argocd.rs
ShmuelOps: Right — fixed in 4400f2e. No advice anymore: the jump uses `switch_context_inner(context, true)` (the reload path), so the same-named-but-remapped context actually reconnects. Covered in `argocd_destination_resolving_to_the_current_context_is_local`.

src/app/argocd.rs
ShmuelOps: Fixed in 4400f2e. `ArgocdReturn` carries the managed row; `reopen_argocd` arms `argocd_reselect`, and the `Msg::Argocd` handler selects `managed_row_at(row)` when the reopened view reports (falling back to the first target as before). Test: `argocd_remote_jump_esc_returns_to_the_row_it_left` — keys arm row 3, the report puts the cursor on `Service/web-3`.

src/ui.rs
ShmuelOps: Agreed — removed in 4400f2e. `src/ui.rs` is untouched vs `main`; `argocd_row_jumps` and the `draw_findings` predicate are gone. The PR is items 1 and 2 only now.

src/ui.rs
ShmuelOps: Moot — the arrow rewiring is removed (see the scope thread).

src/app/argocd.rs
ShmuelOps: Done in 4400f2e — see the thread above: the row travels in `ArgocdReturn` and is applied when the reopened view reports.

ShmuelOps:
Reworked in 4400f2e, rebased on current `main`, still one commit:

- **none reported / summary rows** — no longer map to a resource; the arrow is gone.
- **remapped context name** — the jump reloads through `switch_context_inner(_, true)` instead of advising a `:ctx` that early-returns.
- **esc loses the row** — the managed row travels in `ArgocdReturn` and the reopened view's report selects it (`managed_row_at`).
- **arrow marker** — removed; `src/ui.rs` is identical to `main`. Items 1 and 2 only.
- body and test count updated; the duplicate at ShmuelOps/sofka#1 is closed.

@nklmilojevic — Claude Code with Fable 5.1. The early sloppiness is on me: I pushed before running it against a real hub and let the bot's review drive the iteration. This round I cut the inline comments to five one-liners and kept the doc comments short; happy to trim further if you point at any.

@mesutoezdil thanks for the live run on service-east → apps-dev-east; the row restore is in this push.

src/app/argocd.rs
```diff
@@ -270,12 +290,120 @@ impl App {
...
+        let ArgocdReturn {
+            source, kind, row, ..
+        } = back;
+        self.set_root_view(kind);
+        if let Some(ns) = source.metadata.namespace.clone() {
+            self.namespace = ns;
+        }
+        self.record_history();
+        self.start_watch();
+        self.argocd_reselect = Some(row);
```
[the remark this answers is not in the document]
ShmuelOps: Fixed in 7f31369: `argocd_reselect` is cleared in `cancel_argocd_request` (leaving the view, refresh, source/context change) and on every fresh `show_argocd`; `reopen_argocd` arms it after the open, so only the restored view's own report can consume it. Test: `argocd_reselect_is_dropped_when_the_restored_view_is_left` — `esc` out of the restored view before its report arrives leaves nothing armed.

mesutoezdil:
quick question (before reviewing): could you test all changes in your argocd envs? @ShmuelOps

nklmilojevic (changes requested):
Close, but one thing to fix before merge.

**Failed jump drops the Argo CD view.** `jump_to_remote_managed_resource` sets `self.mode = self.return_mode` before arming the switch. If the switch fails, the `Msg::ContextSwitched` error arm in `lifecycle.rs` clears `pending_argocd_target` and only hands back `pending_argocd_return`. The user ends up on the hub table with no Argo CD view and no way back. An unreachable spoke (VPN down, expired creds) is the most likely failure in a hub/spoke setup, and no test covers it. On that path, reopen the view from `jump.back` (`reopen_argocd`) and add a `handle_key` test for it.

**Minor: the return trip doesn't restore the table.** `reopen_argocd` forces the namespace to the source object's namespace and goes through `set_root_view`, which resets filter and sort. If you started on all-namespaces Applications, you come back scoped to `argocd`. Either keep the original namespace in `ArgocdReturn` or say in the PR that this is intended.

**Minor: comments.** Fewer than last round, but every new struct, field and fn still has a doc comment, some several lines long. Trim them to what isn't obvious from the name.

For the record: merged with current `main` locally, and clippy `-D warnings` plus `cargo test --locked` (1754) pass. Please also cover @mesutoezdil's request to test against a real hub before the next round.

src/ui.rs
ShmuelOps: I took it out, then put it back in 0676e93 after testing on a real hub — with a reason, not by reflex. Upstream already draws `→` on every finding that carries a target, which is every managed row of a *local* Application (the in-cluster apps show it today). Item 1 makes a *remote* managed row go somewhere on `⏎` too; without the mark, the same view shows arrows on local rows and none on remote rows that do the same thing. So this is the existing marker keeping its meaning rather than a new feature — it's why I'd fold it into 1 instead of a later item 5. If you still want it out, say so and it goes; it's one `draw_findings` predicate plus one test.

src/ui.rs
ShmuelOps: Fair — now covered: `argocd_view_renders_the_jump_marker_on_remote_managed_rows` renders the view through `TestBackend` and asserts the `→` cell on remote managed rows, and its absence on the heading, the "… and 1 more" summary, the Source line, and on every row for an unresolved destination. Reverting the `draw_argocd` wiring fails it.

ShmuelOps:
0676e93, after a round on a real hub:

- **`→` is back, as the existing marker keeping its meaning.** Local managed rows already carry it upstream; remote rows now jump too, so they get the same mark. Render-level test through `TestBackend` that fails if reverted. Case made on the thread — happy to drop it if you disagree.
- **`esc` after a local jump now reopens the Argo CD view** on the row it left, like the remote one, instead of popping to the Applications table. The in-cluster case went through `navigate_to_target`'s frame push, so the two paths behaved differently.
- Body updated. Still one commit, on current `main`, 1751 tests.

ShmuelOps:
Manual run of `0676e93` on a real hub and spokes: Argo CD on the mgmt cluster, Applications deploying by `destination.name` `eks-dev-general` / `eks-prod-general`, kubeconfig contexts `mgmt` / `dev` / `prod` (so the names resolve to the aliases through the ARN tail from #570).

**Core path — pass**

1. `⏎` on a managed row of a dev Application: switches to `dev` (the alias, not Argo's spelling), lands on the object's table scoped to its name, flash ends "esc returns to the Application".
2. Same on a prod Application: lands in `prod`.
3. `esc` from the landed table: back on mgmt, Argo CD view reopened on the same Application, cursor on the row I left — tested deep in a long managed-resources list.
4. `esc` order: one press returns from a clean landed table; marks / a typed filter are cleared first, as on any root.
5. A second `esc` lands on the Applications table with the Application selected.
6. `Tab` through pods / deployments / … in the spoke, then `esc` at the root: still returns to the Argo CD view on mgmt.
7. `:pods` in the spoke, then `esc`: ordinary root `esc`, no switch — the way back is dropped by an explicit query.
8. Jump to a Deployment row, `⏎` again into its pods, `esc` twice: first pops to the scoped table, second returns to mgmt.

**Found while testing, fixed in the same push**

- **In-cluster Applications: `esc` didn't come back.** On an Application with destination "this cluster", `⏎` on a managed row opened the object but `esc` popped to the Applications table instead of the Argo CD view — the local path went through `navigate_to_target`'s frame push while the remote path used the root-view landing. Both now share the landing, so `esc` reopens the view on the row left in either case (`argocd_local_jump_esc_returns_to_the_view`).
- **No `→` on remote managed rows.** Local managed rows already carry the marker upstream (they have targets); after this change remote rows jump too, so they were the only rows that go somewhere on `⏎` without saying so. Restored with a `TestBackend` render test (`argocd_view_renders_the_jump_marker_on_remote_managed_rows`) — see the thread for the scope discussion.

ShmuelOps:
50ec316, rebased on current `main` (1765 tests), one commit:

- **Failed jump kept the view.** A switch to the spoke that fails now reopens the Argo CD view on the row it left, then shows the failure; before, the user was dropped on the hub table with no way back. Test: `argocd_remote_jump_failed_switch_reopens_the_view`.
- **Return restores the table.** Namespace and filter travel in the return state and come back as left (all-namespaces stays all-namespaces). Sort resets, as any fresh root view does — stated in the body.
- **Landing sets `Mode::Table`.** Both jump paths stopped relying on `return_mode`, which showed Xray/Explain/Adjacent instead of the object when the view was opened from there (Greptile's finding). Test: `argocd_jump_lands_in_the_table_whatever_the_view_was_opened_from`.
- **Comments.** Every new doc comment is one line or gone; the structs also moved off the `PREFIX_SCAN_*` doc comment they had split.

Real-hub testing: the core path (1–8) is reported above on the previous commit; the edge cases follow on this one.

src/app/argocd.rs
```diff
@@ -251,16 +275,22 @@ impl App {
...
                     .selected()
                     .and_then(|i| self.argocd_items.get(i));
                 match selected.and_then(|f| f.target.clone()) {
+                    // A managed row opens the same way here as in another
+                    // cluster, so `esc` comes back to this view either way.
+                    Some(_) if self.managed_jump_here().is_some() => {
+                        let jump = self.managed_jump_here().unwrap();
+                        self.mode = self.return_mode;
+                        let context = self.cluster.context.clone();
+                        self.open_remote_managed_resource(jump, &context);
```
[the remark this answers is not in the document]
ShmuelOps: Fixed in 50ec316: `open_remote_managed_resource` sets `Mode::Table` itself and neither jump path uses `return_mode` anymore, so a view opened from Xray/Explain/Adjacent lands on the object's table. Test: `argocd_jump_lands_in_the_table_whatever_the_view_was_opened_from` (sets `return_mode = Xray`, expects `Table` after `⏎`).

src/app/argocd.rs
```diff
@@ -270,12 +298,129 @@ impl App {
...
+            filter,
+            row,
+            ..
+        } = back;
+        self.set_root_view(kind);
+        self.namespace = namespace;
+        self.filter = filter;
+        self.record_history();
+        self.start_watch();
+        self.show_argocd(source);
```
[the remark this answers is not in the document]
ShmuelOps: Fixed in e85bdcb: the mode the view was opened from travels in `ArgocdReturn` (`return_mode`) and `reopen_argocd` restores it after `show_argocd`, so `esc` out of the reopened view goes back to Xray / Explain / Adjacent as before the jump. Their state is not cleared by a switch, so there is something to return to. Extended `argocd_jump_lands_in_the_table_whatever_the_view_was_opened_from`: `⏎` → Table, `esc` → Argo CD view with `return_mode == Xray`, `esc` → Xray.

